PR #21, opened 2018-12-27
Make it simpler to write custom principals, and fix a typo in JWTAuthContextInfoProvider

sberyozkin:


sberyozkin:
Hi Scott, it is a basic refactoring, pushing some of the common code to the abstract class:
* reuse toString() and implies() implementations
* reuse the way the principal name is calculated
* push down some of the type conversion utility code
* the code dealing with the raw token is also in the abstract class 

Now the default implementation is easier to read and the KC one looks like this:

https://github.com/thorntail/thorntail/compare/master...sberyozkin:THORN-1852?expand=1#diff-33bad9870eec34c2d1d35bb4552621a1
Please review

sberyozkin:
I moved the code for loading the Json Web Keys over HTTPS from JWTContextInfo which is really just a collection of the configuration properties to the default factory, for the custom ones to avoid having to include jose4j dep.

PR #44, opened 2019-05-02
Log validation error

sberyozkin:


implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -95,6 +95,7 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
             
             return jwtContext;
         } catch (InvalidJwtException e) {
+        	logger.warnf("Token is invalid: %s", e.getMessage());
```
Ladicek: tab indentation :-)

sberyozkin:
@Ladicek,  @starksm64, sorry, missed it completely, might be to do with me creating a new smallrye-jwt workspace, will reconfigure it. thanks for fixing this tab issue :-)

PR #55, opened 2019-05-15
Prevent NPE in JWTAuthFilter and handle other token locations

MikeEdgar:
This fixes a potential `NullPointerException` situation and adds support for cookie and custom HTTP header bearer tokens to `JWTAuthFilter`. There are several test cases added here along with two `test` scope Maven dependencies to support them.

* Replace System.err.printf statements with logging and avoid printing the token to the log.
* Add test cases for JWTAuthFilter#getBearerToken and add two test dependencies (Mockito and Jersey to support tests).

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -71,6 +71,24 @@
       <groupId>org.jboss.logging</groupId>
       <artifactId>jboss-logging</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.27.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.core</groupId>
+      <artifactId>jersey-common</artifactId>
+      <version>2.27</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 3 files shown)


implementation/pom.xml
```diff
@@ -71,6 +71,24 @@
...
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.27.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
```
sberyozkin: Is this dependency needed even at the test level ?

sberyozkin:
Looks good in general, thanks

implementation/pom.xml
MikeEdgar: @sberyozkin - I used Mockito directly in the tests added in this PR. If there is another preferred approach for unit tests, I'm happy to rework them.

sberyozkin: Thanks, can you please give RestAssured a try ? It is the preferred framework for testing in Quarkus/Thorntail so ideally we'd avoid bringing the JAX-RS implementation specific test dependencies without actually doing the JAX-RS centric tests. Not a big deal I guess if the Jersey test dependencies will be brought in, but the RestAssured way might be better.
@kenfinnigan @starksm64 Hi Ken, Scott, please comment if you have other preferences
thanks

sberyozkin: https://blog.sysco.no/automatic/tests/web-service-testing/ is one place where the mocks are described

MikeEdgar: @sberyozkin - the particular tests added aren't against JAX-RS endpoints, but rather against the new package private method added to the filter. This particular Mockito library isn't specific to the JAX-RS implementation. Thoughts?

sberyozkin: Can you please remove the jersey-core dependency ? I've verified that your test still passes (since it is probably pulled in by Mockito anyway), could not push this minor update to your branch

sberyozkin: Never mind, I can take care of it with my next PR :-)

MikeEdgar: @sberyozkin - sorry about this one. I'm on the same page now regarding the Jersey dependency.

PR #65, opened 2019-05-24
Add an error message when a public key is loaded as a resource using mp.jwt.verify.publickey.location

ekans:
The imports order have changed because I'm using the default formatter of IntelliJ.
I can fix that if you want but I don't know the rules.


sberyozkin:
@ekans thanks for the PR, we don't have a specific import order enforcement yet, so we can adjust later on once/if we agree on some specific one

Issue #66, opened 2019-05-24
KeyLocationResolver base64 decoding breaks pem processing


sberyozkin:
As part of my earlier PR #61  I tried to optimize the way base64 processing is done and pushed it out of JWK processing code to the loadContent ones. This broke the PEM processing which I only caught with Thorntail TCK tests.
I propose simply remove the Base64 decoding attempt as it was only meant originally for JWK but JWKs are not usually get Base64 encoded as the whole (but Base64Url encoded if they included as JOSE header properties).

PR #67, opened 2019-05-27
Remove Base64 decoder code from KeyLocationResolver

sberyozkin:


sberyozkin:
Hi @starksm64 

I've removed the base64 decoder code and Thorntail TCK is good. It was only originally done for JWKs but as I've mentioned at #66 (the issue was caused by me not testing it well), JWK(s) is usually Base64URL encoded, at least if it goes as a Jose header property.
Have also improved the JWK set related test, it was not clear before whether the expected failure was caused by a mismatch or by the fact I had one of the keys in the set being invalid

PR #69, opened 2019-05-30
Remove the 2nd JWTAuthContextInfoProvider constructor

sberyozkin:


sberyozkin:
I added it earlier for the test purposes, but Quarkus is not happy about it, so I've just introduced the setters instead

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -198,18 +188,34 @@ protected void setTokenHeadersAndGroups(JWTAuthContextInfo contextInfo) {
         return mpJwtPublicKey;
     }
 
+    public void setMpJwtPublicKey(Optional<String> mpJwtPublicKey) {
```
Ladicek: Would it be possible to make these methods package-private and move the test to the same package? Alternatively, add a comment that these methods are only exposed for test purposes.

I'm not sure if this could expose some security hole...

sberyozkin: @Ladicek  Yes, it is not really cool. Right now the test which depends on sits in a different module which depends on the tck key. So I'll introduce few static methods to keep Qurakus happy and avoid making it mutable, thanks

sberyozkin: @Ladicek hopefully it is better now, let me know please how it looks to you

Ladicek: Yea, I think that's much better, thanks!

sberyozkin:
@Ladicek thanks

PR #70, opened 2019-06-11
Throw the exception with the useful message if context is null

sberyozkin:


sberyozkin:
This addresses https://github.com/quarkusio/quarkus/issues/2725

PR #72, opened 2019-06-12
Removing the monolitic header/groups function and updating the error message

sberyozkin:
The original function I added was a quick attempt to have it reused from Thorntail but the code for setting the cookies is not reusable. The error message in the non-optional producer method is more user friendly as well (per @starksm64 suggestion)


Ladicek:
LGTM, but I'm not exactly fond of how we're changing public API quite often. (The API here, AFAIK, is not for direct consumption, but for integrators, so it's not _that_ big of a deal, but still.)

sberyozkin:
Agreed that more care should be taken now that smallrye-jwt is becoming much more visible, thanks

PR #74, opened 2019-06-13
Support the custom groups path

sberyozkin:


implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -104,6 +109,53 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
 
     }
     
+    private List<String> checkGroupsPath(JWTAuthContextInfo authContextInfo, JwtClaims claimsSet) {
+        if (authContextInfo.getGroupsPath() != null) {
+            String[] pathSegments = authContextInfo.getGroupsPath().split("/");
+            if (MAX_ROLES_PATH_DEPTH < pathSegments.length) {
+                logger.warnf("Roles path depth is too large: %d, maximum depth is %d",
```
Ladicek: Should be an ERROR in my opinion, and it should say that it's ignoring the custom configuration completely. In fact, would it be possible to verify the custom path _at deployment time_ and log an error early?

sberyozkin: `JWTAuthContextInfoProvider` can check it earlier. Let me have a look

Ladicek:
I'm not exactly sure about the custom query language, but at the point of evaluation, it's no longer JSON, so unfortunately we can't use JSON Pointer. Sigh.

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
sberyozkin: @Ladicek I moved this code to the utility class (I've been planning to introduce it for a while), so that we can reuse this code (and the token cookie code, etc) from Thorntail (where META-INF resources are checked which is done outside of `JWTAuthContextInfoProvider`). What do you think ?

sberyozkin:
Yes, it is only a Map which is what the claims set is after the token has been validated

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
Ladicek: Makes sense to me. If I understand correctly, you intend to do a bigger refactoring, so I'd perhaps add a warning to these new classes that public API is not guaranteed, or something like that. Or just bump major versions tirelessly :-)

sberyozkin: @Ladicek the next release will be a new artifact id so it would be good to start from 1.0.0 I guess. So far the only user of these protected helpers is Thorntail (AFAIK). But indeed the time where I can keep moving functions around is coming to the end :-). I'll send an email to the smallrye forum to check just in case before a release

Ladicek: Actually `smallrye-jwt-1.1` version `1.0.0` was already released, so ... Well I don't mind `2.0.0` at all :-)

sberyozkin: @Ladicek This is a surprise, I did not know it was released yesterday. OK, I don't mind either 2.0.0, I'll send an email to the forum when the next release is due

Ladicek:
I'd just like to note that the query language should be documented wherever it's exposed. It's not really "JSON path".

PR #76, opened 2019-06-18
Applying the formatter/import order changes after the smallrye parent upgrade

sberyozkin:


tck/src/test/java/KeyLocationResolverTest.java
```diff
@@ -15,8 +15,23 @@
  *   limitations under the License.
  *
  */
-
-
+/*
```
kenfinnigan: Is the license duplicated here?

tck/src/test/java/TestTokenClaimTypes.java
```diff
@@ -17,19 +17,28 @@
...
-import io.smallrye.jwt.auth.principal.JWTAuthContextInfo;
-import io.smallrye.jwt.auth.principal.JWTCallerPrincipalFactory;
-import org.eclipse.microprofile.jwt.Claims;
-import org.eclipse.microprofile.jwt.JsonWebToken;
-import org.eclipse.microprofile.jwt.tck.util.TokenUtils;
-import org.jboss.arquillian.testng.Arquillian;
-import org.testng.Assert;
-import org.testng.annotations.BeforeClass;
-import org.testng.annotations.Test;
+/*
```
kenfinnigan: duplicated license?

tck/src/test/java/TestTokenWithGroupsPath.java
```diff
@@ -1,3 +1,22 @@
+/*
```
kenfinnigan: I think there are quite a few duplicate licenses in the files.

Can you check them?

sberyozkin: Sure will do

sberyozkin: @kenfinnigan looks like the default package was causing it so I moved those test classes into a non empty package; also updated all the license headers to have 2019 and starting with a single asterisk

PR #78, opened 2019-06-19
Added configurations to better control JWT validations-

radcortez:
I've added a few configurations to allow for a better control on the JWT validation.

First, there is some code that allows to load JWKS using jose4j, but that code is not reachable `(io.smallrye.jwt.auth.principal.DefaultJWTTokenParser#loadJsonWebKeys)`, since its uses the jwksUri, that is setted using the mp config `mp.jwt.verify.publickey.location`, but setting that config also triggers the `followMpJwt11Rules` which uses the KeyLocationResolver for MP and not the JwksVerificationKeyResolver.

Then, I've also added some configurations to allow whitelisting other algorithms, supported by jose4j, and to turn on / off some of the required checks.

Of course, the defaults are the expected ones of MP JWT, and the behaviour is only changed by using specific smallrye configurations.

Let me know if this is ok. I'll be happy to do any other changes if required.


sberyozkin:
@radcortez Thanks for the PR.
I'm not sure about making either the subject or expiration claims optional, both are essential, why would you like to make them optional ?

radcortez:
Build is failing, but seems to be an issue with the Maven Repo:

```
[ERROR] Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.8.1:format (default) on project smallrye-jwt-parent-1.1: Execution default of goal net.revelc.code.formatter:formatter-maven-plugin:2.8.1:format failed: Plugin net.revelc.code.formatter:formatter-maven-plugin:2.8.1 or one of its dependencies could not be resolved: Could not transfer artifact net.revelc.code.formatter:jsdt-core:jar:2.6.0 from/to central (https://repo.maven.apache.org/maven2): GET request of: net/revelc/code/formatter/jsdt-core/2.6.0/jsdt-core-2.6.0.jar from central failed: Premature end of Content-Length delimited message body (expected: 28356165; received: 10745595 -> [Help 1]
```

I executed the same command on my box and it seems to run fine.

radcortez:
> I'm not sure about making either the subject or expiration claims optional, both are essential, why would you like to make them optional ?

Well it is only a configuration. It will be on by default. Unfortunately, I'm dealing with an in-house JWT provider that does not include the Subject, so I get an invalid token. I understand that this is not a good reason, but I'm I'm unable to configure it, I won't be able to use this implementation to validate my JWT's.

> And please clarify why did you prefer to drop the filtering our of all but RS256 keys - would you like to use some other algorithm ?

The algorithms are already whitelisted on jose4j config, so I believe we don't need additional filtering when reading the keys from the JWKS. Just read everything and let jose4j whitelisting do the filtering, or we will need to double filter between jose4j and reading the keys from the JWKS.

Hope it makes sense.

sberyozkin:
Sure...
Do these tokens include either `upn` or `preferred_user_name` ? If not and if no `sub` is available then the user can not even get the principal name.
And these tokens do not include `exp`, right ? The concern is we are letting the time uncontstrained tokens in, so there is a no protection at all at the smallrye-jwt level then against the leaked/stolen access tokens without `exp`.
Can you also please clarify further what you mean the algorithms are already whitelisted by Jose4j. Please share a link to the Jose4j source code.

Thanks

radcortez:
> Do these tokens include either `upn` or `preferred_user_name` ? If not and if no `sub` is available then the user can not even get the principal name.

No. They use a custom named scope to store the Principal Name. That is another story. I'll just retrieve the name from the custom claim, but yes I won't have Principal support.

> And these tokens do not include `exp`, right ? The concern is we are letting the time uncontstrained tokens in, so there is a no protection at all at the smallrye-jwt level then against the leaked/stolen access tokens without `exp`.

They do. Since I was making other stuff configurable, I also made that configurable, but we can remove that piece.

> Can you also please clarify further what you mean the algorithms are already whitelisted by Jose4j. Please share a link to the Jose4j source code.

Yes, that is done here:
https://github.com/radcortez/smallrye-jwt/blob/d1090e873434ac0bbb156728b9d54cd7843b6b95/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java#L63-L71

> 
> Thanks

radcortez:
> FYI, one can register a custom verification factory, we do in Thorntail for example

Can you point me at some code. I was looking to provide my custom implementation of io.smallrye.jwt.auth.principal.DefaultJWTTokenParser, but it didn't look that easy.

sberyozkin:
I wonder, for both the missing `sub` claim, should we rather go the default values path, the same what we did for `groups`. For example, we have `smallrye.jwt.claims.groups`. We can have `smallrye.jwt.claims.sub`=`microprofile-jwt-client`, something like that.

radcortez:
Yes, we could do something like that, but we will still need to disable the Sub validation. We can remove the specific config to do it and make it implicit if you set a `smallrye.jwt.claims.sub`.

sberyozkin:
Thanks again. Just few days ago we added a support for mapping custom roles claim to the `groups`. So one more options, add a property `smallrye,jwt.sub,path` and tackle it similarly to the groups 1) check if the custom path exists, if yes, map to to `sub` otherwise use the default value ? 
It will make these custom tokens meet the typical expectations, what do you think ?

Yes, I'd prefer the default implementation not to have even accidental paths to let the open-ended tokens, happy you don't mind as well, please drop that bit.

Re the whitelisting, sorry, I got confused, see in your code RS256 is still default

sberyozkin:
Sure with either the default or mapped value we will guarantee the `sub` wil be available so it would be fine just delete the subject validation request

radcortez:
Ok, thanks for all your input. Let me do the proposed changes and I'll submit and update to the PR. 

Cheers!

sberyozkin:
Sounds good, thanks (will go offline shortly till tomorrow). 
FYI, given your initial PR, I thought about opening 2 new issues (generalize the required claims constraint such that many of them can be listed with a single property) and make the default parser easier to extend (at the moment the custom validation is only possible by providing a custom factory)
Cheers

sberyozkin:
@radcortez if the `smallrye.jwt.sub.path` works for you then can you please make `path` a 3rd segment: `smallrye.jwt.path.sub`, and I will do separately the same for `smallrye.jwt.path.groups`, so that the users can group the paths to more than one claim under the same top level `path` YAML prefix

radcortez:
@sberyozkin I had some work on it yesterday, just to support a simple sub mapping. I was ignoring path support for now. Do you want me to also implement support for it?

sberyozkin:
@radcortez in your case, is it sufficient to point to a top level custom claim value to get the subject value ? Can you type in a custom token format (only a relevant part, change the actual custom subject claim name if it can be sensitive).
Let me also comment on the specific PR parts

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -53,11 +53,22 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
...
                     .setRequireExpirationTime()
-                    .setRequireSubject()
-                    .setSkipDefaultAudienceValidation()
-                    .setJwsAlgorithmConstraints(
-                            new AlgorithmConstraints(AlgorithmConstraints.ConstraintType.WHITELIST,
-                                    AlgorithmIdentifiers.RSA_USING_SHA256));
+                    .setSkipDefaultAudienceValidation();
+
+            if (authContextInfo.getDefaultSubjectClaim() == null ||
+                    Claims.sub.name().equals(authContextInfo.getDefaultSubjectClaim())) {
```
sberyozkin: I think we can drop the check for `sub`, it does not save us much

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -53,11 +53,22 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
...
+                builder.setRequireSubject();
+            }
+
+            if (authContextInfo.getWhitelistAlgorithms().isEmpty()) {
+                builder.setJwsAlgorithmConstraints(
+                        new AlgorithmConstraints(AlgorithmConstraints.ConstraintType.WHITELIST,
+                                AlgorithmIdentifiers.RSA_USING_SHA256));
+            } else {
+                builder.setJwsAlgorithmConstraints(
```
sberyozkin: I'd like to suggest to do more work around it:

- please use `AlgorithmIdentifiers.valueOf(String).name()` or similar to ensure these are valid values
- please filter out the symmetric algorithms [starting from HS](https://bitbucket.org/b_c/jose4j/src/d05e68f9822cd479f2d42edfa5436d3d749ae3aa/src/main/java/org/jose4j/jws/AlgorithmIdentifiers.java#lines-25). As I've mentioned earlier MP JWT does not like the fact the secret key has to be visible to both the client and all the microservices nodes which does not scale and is problematic from the key management POV.

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -86,6 +97,10 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
             JwtClaims claimsSet = jwtContext.getJwtClaims();
 
             claimsSet.setClaim(Claims.raw_token.name(), token);
+            if (authContextInfo.getDefaultSubjectClaim() != null) {
```
sberyozkin: Please update the condition to start from the `!claimsSet.hasClaim(Claims.sub.name())`

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -86,6 +97,10 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
...
+                claimsSet.setClaim(Claims.sub.name(),
+                        claimsSet.getClaimValue(authContextInfo.getDefaultSubjectClaim()));
+            }
```
sberyozkin: What I was suggesting earlier is to fail at this point if after all the checks `sub` is still null because my understanding is MP JWT, if it accepts the token. should be able to offer a valid (named) `Principal`, so that if `upn` or `preferred_username` are also missing then `sub` can be used as a last resource.
@starksm64, what do you think ?

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -183,11 +213,29 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
+
+        if (whitelistAlgorithms != null && !whitelistAlgorithms.isEmpty()) {
+            contextInfo.getWhitelistAlgorithms().addAll(whitelistAlgorithms);
+        }
+
+        // JWKS based on jose4j
+        if (jwksLocation != null && jwksLocation.isPresent()) {
+            contextInfo.setJwksUri(jwksLocation.get());
+            contextInfo.setJwksRefreshInterval(jwksRefreshInterval);
+            contextInfo.setFollowMpJwt11Rules(false);
```
sberyozkin: Good catch, I'll need to check what is going on in Thorntail because that code is reachable from thorntail and is being tested

radcortez:
> @radcortez in your case, is it sufficient to point to a top level custom claim value to get the subject value ? Can you type in a custom token format (only a relevant part, change the actual custom subject claim name if it can be sensitive).

Yes it is sufficient. 

Sorry, I didn't understand the second part. What do you mean exactly?

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
radcortez: Ok

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
radcortez: Yes, I was still not done with this. 

I was planning to add a setWhitelistAlgorithms method to SmallryeJwtUtils and perform validation regarding the accepted algorithms types and output a log message if you are setting an unsupported algorithm.

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
radcortez: Ok.

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
radcortez: Well, if we are using the default combination of sub and setRequireSubjectm jose4j will validate that for us. If that is not the case, then maybe we should implement a custom Validation so we do the additional checks on `ups` and `preferred_username` and let jose4j validate that as well.

sberyozkin: @radcortez or may be we can reintroduce your suggested required.subject property and fail now only if sub is still null and the user has explicitly said it is not really needed, just need to check with @starksm64

sberyozkin:
@radcortez OK, if it is enough for your case then it is good. I just wanted to see the example in case it was not sufficient to evaluate how realistic it was to get to such a claim, example, if it were nested somewhere, etc... We can revisit it later if needed

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
radcortez: Ok. Cool. If that is the case, then we only need to play with setRequireSubject in jose4j

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Maybe you have another bean with a Specializes or Alternative to change the value? As is, the JwksVerificationKeyResolver is not reachable. I'm using Quarkus btw.

radcortez:
> @radcortez OK, if it is enough for your case then it is good. I just wanted to see the example in case it was not sufficient to evaluate how realistic it was to get to such a claim, example, if it were nested somewhere, etc... We can revisit it later if needed

Wonderful, so we can leave it as is, right?

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
sberyozkin: i'd like to keep it simple, the early `sub` validation is a problem in your case, so this PR removes it, and tries to get it from some other custom claim (the user may set it to `sub` but we are not really doing this custom claim check just to delay the default `sub` validation), but in the end we need to either fail if the `sub` (possibly mapped from the custom claim) is still null or let it go if the users allows for it. 
Now that I think about it, it is really a bigger issue, whether to fail the verification if no named Principal can be created (using one of those 3 claims to get the name), so I'd like to propose to tackle it separately (possibly with the help of the custom Validator as per your suggestion and some specific property). As such I'd like to suggest just to throw the error at this point

radcortez: I understand. Let me have a look and I'll implement something to always fail if is not able to find a sub claim either on `sub` of the configured property. Let's also check `ups` and `preferred _username`.

sberyozkin:
@radcortez I apologize, I got confused, I've looked at your code again.
The way `smallrye.jwt.claims.subject` is processed is different to the way  `smallrye.jwt.claims.groups` is processed and they have to be consistent, in both cases these properties point to the default value, i.e the value to use when no relevant info exists in the token.
So, `smallrye.jwt.path.subject` will help because you'd just set it to "a" or "a/b" and it will work. Simply copy and paste the code for checking the path to the groups and and only replace the code which checks the final value is a simple property.
Then you'd just check the subject exactly how the groups is checked (except that for now please throw the exception if sub is null after all, as pointed above and we will deal with the whole principal name story afterwards)
Hope you are fine with all of it, lots of comments from me but your PR has raised some interesting issues and it is nearly there, many thanks

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: Yes, something like that

sberyozkin:
@radcortez FYI, going offline for a bit but will continue with finalizing the review later; thanks for your effort so far

radcortez:
@sberyozkin No worries, thank you for all the help.

I'm actually busy with other stuff, so most likely I'm only going to be able to pick this up later at the day.

Cheers!

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
sberyozkin: I'd like to have a different issue opened to have a dedicated property (true by default) that if neither of these 3 claims is available then fail. I'm Ok though with it being done as part of this PR, the only question is how to name this property, `require.name.claim` or something like that ?

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: The more I'm looking at this the more I'm uncomfortable (not with your code :-) ) but with the way it is currently done in smalrye-jwt (if jwks happens to be on the classpath then they are loaded from one resolver, if it is https uri - then it is done inside the parser) and then this follow the rules property which is confusing as well.
Please drop this code for now, I will open an issue to consolidate (I'd rather do it then making this separation more visible which would be the case with the new properties you propose) and will link to it below. Keep the `jwksRefreshInterval` property though please, we can update the docs later to say it is ignored if jwks are not on the classpath.

radcortez: I see. Yes, I understand your point :)

The`jwksRefreshInterval` is only being applied in the http KeyResolver and not in the class path resolver, which may be more confusing. That was the reason why I wanted to use the http KeyResolver.

Maybe we could use the jose4j HttpsJwks and implement our own SimpleGet to also support loading JWKS from the classpath, and basically drop the current code used to load JWKS and use the jose4j one.

sberyozkin: https://github.com/smallrye/smallrye-jwt/issues/82.

Let me summarize all below, thanks

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
sberyozkin: Ignore the last point now please, summarizing below

radcortez: Ok. Thanks. It is becoming somehow confusing with so many parallel comments :)

sberyozkin:
Hi @radcortez 
Thanks again for your PR, ideas and patience so far :-)
Let me summarize all the various points here:
1. PR removes the early 'sub' validation, and I proposed to introduce `smallrye.jwt.path.sub` and `smallrye.jwt.claims.sub` properties to deal with the missing `sub` very similarly to the way `groups` is handled. (FYI, at the moment we have `smallrye.jwt.groups.path` for the path to the custom `groups`  but [it will be changed](https://github.com/smallrye/smallrye-jwt/issues/81)). 
2. To ensure the named principal is available the proposal is to introduce `smallrye.jwt.require.named-principal`: if this property is set (default) then the code will check that at least one from `upn` or `preferred_user_name` or `sub` is not null (after `sub` is possibly mapped as part of the point 1 above)
3. Some minor cleanup is expected for the whitelist property
4. Please drop all the JWKS related updates but keep the jwks expiration grace property, I will take of the consolidation in https://github.com/smallrye/smallrye-jwt/issues/82 

Hope you are Ok with the above, lets try to finalize your PR by Monday eve, and I'll take care of #82 and #81 and do  a 2.0.0 release.
Enjoy the weekends
Sergey

radcortez:
@sberyozkin Sounds good. Thanks for the summary.

I'll work on that. If you like I can also have a look into #81 and #82.

sberyozkin:
@radcortez Thanks, I'll need to coordinate with the Thorntail code for #82 (it will break some code there :-)), but if you will find some extra time to get #81 fixed as part of this PR then it would be nice. We can also discuss further improvement ideas you may have next (open new issues or send an email to the smallrye list) and I/we will definitely welcome your future contributions :-).

radcortez:
@sberyozkin Please have a look into the update PR and tell me what you think.

I haven't done anything in #81, because I want to discuss something with you first. 

I've implementes sub path using the same logic applied to groups. Looking into the configurations available in groups (now also equivalent in sub), `smallrye.jwt.claims.groups` and `smallrye.jwt.groups.path` I feel they are somehow redundant, since one uses a single property name (like a path with 1 deep) and the other one a composed path (with max path 4).

I think we should drop one of them, or if we want to support both, `smallrye.jwt.groups.path` should work as an actual path (without the claim name) and `smallrye.jwt.claims.groups` will be appended in the end of the path. What do you think?

sberyozkin:
`smallrye.jwt.claims.groups` is not a path but a default value, if no `groups.claim` is available at all. Similarly to 'sub'.

implementation/src/main/java/io/smallrye/jwt/SmallryeJwtUtils.java
```diff
@@ -16,26 +16,49 @@
...
 import org.jboss.logging.Logger;
+import org.jose4j.jws.AlgorithmIdentifiers;
 
 import io.smallrye.jwt.auth.principal.JWTAuthContextInfo;
 
 /**
  * Utility methods for dealing with decoding public and private keys resources
  */
 public class SmallryeJwtUtils {
+    private static final Integer MAX_SUB_PATH_DEPTH = 4;
```
sberyozkin: Let have a single `MAX_PATH_DEPTH` property

implementation/src/main/java/io/smallrye/jwt/auth/principal/ClaimSubPathResolver.java
```diff
@@ -0,0 +1,53 @@
...
+            if (claimValue instanceof Map) {
+                @SuppressWarnings("unchecked")
+                Map<String, Object> nextMap = (Map<String, Object>) claimValue;
+                int nextStep = step + 1;
+                return findSub(authContextInfo, nextMap, pathArray, nextStep);
+            } else {
+                logger.warnf("Claim value at the path %s is not a json object", authContextInfo.getGroupsPath());
+            }
+        } else if (claimValue instanceof String) {
+            // last segment
```
sberyozkin: Cast to String is enough without try/catch

implementation/src/main/java/io/smallrye/jwt/auth/principal/ClaimSubPathResolver.java
```diff
@@ -0,0 +1,53 @@
...
+            try {
+                return (String) claimValue;
+            } catch (ClassCastException e) {
+                logger.warnf("Claim value at the path %s is not an array of strings", authContextInfo.getGroupsPath());
+            }
+        } else {
+            // last segment
+            logger.warnf("Claim value at the path %s is not an array", authContextInfo.getGroupsPath());
```
sberyozkin: Please update to `is not a string`

implementation/src/main/java/io/smallrye/jwt/auth/principal/ClaimSubPathResolver.java
```diff
@@ -0,0 +1,53 @@
+package io.smallrye.jwt.auth.principal;
```
sberyozkin: I'd like to avoid introducing a dedicated class for checking sub paths only, while the `groups` are checked inside the parser, lets keep it inside the parser code for the moment. Good idea to try to push it outside,  but we can attempt to generalize later so that a single code is dealing will all the claim paths, with the only custom bit being checking the final segment, but we can take care of it later

radcortez:
> `smallrye.jwt.claims.groups` is not a path but a default value, if no `groups` claim is available at all, either directly or at the `smallrye.jwt.path.groups` path. `sub` should be handled the same way

I see. The description for the property is misleading to me. I did indeed see the code that was assigning the config value directly to the claim, but I thought it was a bug.

Do you think if the property changes to something like `smallrye.jwt.claims.groups.value` (added the value) and update the description it becomes more clear? Or do you want to keep it as is?

implementation/src/main/java/io/smallrye/jwt/auth/principal/CustomSubValidator.java
```diff
@@ -0,0 +1,53 @@
...
+                    .map(jwtClaims::getClaimValue)
+                    .anyMatch(Objects::nonNull);
+            if (!hasDefaultPrincipalClaim) {
+                return new Error(ErrorCodes.SUBJECT_MISSING, "No Subject claim is present.");
+            }
+        }
+
+        return null;
+    }
+}
```
sberyozkin: Another good idea but at the moment we tend to keep the various checks inside the parser instead of having some inside the parser and some branching into Jose4J handlers. May be we can push all to the Jose4J eventually but again, I'd like to keep this PR simple, and we can consider the error checks consolidation improvements later. 
By the way the logic does not appear to be correct to me.
I propose push it back to the default parser, and there we do not check the paths or defaults, the sub path/default value should be taken care of earlier similarly to the groups. The check should only be about checking that either `upn` or `preferred_username` or `sub` exist (in this order) and only throw the exception if not

sberyozkin:
@radcortez 

`value` is implicit with YAML, but indeed please update the description. It currently says `Default group name. This property can be used to support the JWT tokens without a 'groups' claim.` May be the 1st part to `Default groups claim value....`. I did mention it earlier but there are many comments now so sorry I did not make it clearer, thanks

implementation/src/main/java/io/smallrye/jwt/auth/principal/CustomSubValidator.java
radcortez: Ok

implementation/src/main/java/io/smallrye/jwt/auth/principal/ClaimSubPathResolver.java
radcortez: This was because I needed the same logic to retrieve the value in the jose4j validator.

radcortez:
@sberyozkin check if it is better now :)

sberyozkin:
@radcortez Looks perfect to me now :-) I'm going to check how it all goes against Thorntail/Quarkus and will merge. Thanks a million for the effort/patience.

(If you'd like, please consider 1) extracting the common code from both sub and groups path setters in the `SmalryeJwtUtils` to an internal one and 2) have the common findPath method in the parser, it would return the final segment object without the casts, while the group and sub logic will cast to either List Or String - I can take care of it later on or may be you might want to do it in the next PR, it is a minor and non urgent refactoring)

Cheers

radcortez:
> @radcortez Looks perfect to me now :-) I'm going to check how it all goes against Thorntail/Quarkus and will merge. Thanks a million for the effort/patience.
> 
> (If you'd like, please consider 1) extracting the common code from both sub and groups path setters in the `SmalryeJwtUtils` to an internal one and 2) have the common findPath method in the parser, it would return the final segment object without the casts, while the group and sub logic will cast to either List Or String - I can take care of it later on or may be you might want to do it in the next PR, it is a minor and non urgent refactoring)
> 
> Cheers

Wonderful. I'll do it in a separate PR. 

It was a pleasure to contribute. I hope I can contribute more stuff in the future :)

sberyozkin:
Cool, glad to hear it, thanks again

sberyozkin:
@radcortez can you please sync your PR against the master ? The command line instructions produce a merge editor which suggests there is a minor issue there (may not be a problem really)

radcortez:
@sberyozkin done. BTW, do you need any help with #82? I also need this for my case :)

sberyozkin:
@radcortez thanks, the tests are still failing, it is something small that has changed, let me try to trace it down.
I did a quick [code](https://github.com/sberyozkin/smallrye-jwt/tree/process_jwks) for #82 yesterday, but I'll need to play a bit more with it depending how Thorntail will 'like' it :-). It will be in the next release for sure, thanks.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -120,9 +121,30 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
     private Optional<String> tokenCookie;
-
     /**
-     * Default group name. This property can be used to support the JWT tokens without a 'groups' claim.
+     * Check that the JWT has at least one of 'sub', 'upn' or 'preferred_user_name' set. If not the JWT validation will
+     * fail. If the properties 'smallrye.jwt.claims.sub' or 'smallrye.jwt.sub.path' are set it will use these values
+     * to validate instead.
+     */
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.require.named-principal", defaultValue = "true")
```
sberyozkin: Hi @radcortez While I'm chasing the issue at the Thorntail side, can you please change the default value to `false` as indeed there could be some custom tokens around without those 3 claims. One would use `true` only to ensure that the application code can always rely on Java principal name

radcortez: Ok.

radcortez: @sberyozkin done!

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -135,6 +157,16 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.expiration.grace", defaultValue = "60")
+    private Optional<Integer> expGracePeriodSecs;
+    /**
+     * List of algorithms to whitelist JWT validation based on jose4j algorithms
+     * list org.jose4j.jws.AlgorithmIdentifiers.
+     */
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.whitelist.algorithms")
+    private List<String> whitelistAlgorithms;
```
sberyozkin: This is what was causing a failure - the property needs to be Optional, please update

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: Thanks; one more thing, please update the docs of this property, remove the 2nd sentence starting from 'If'. And after you also update the whitelist property to Optional I'll be ready to merge I hope :-)

sberyozkin:
All Thorntail tests are passing after fixing the whitelist issue locally

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Done :)

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Done :)

radcortez:
> All Thorntail tests are passing after fixing the whitelist issue locally

Cool. Check if everything is ok now :)

Issue #82, opened 2019-06-21
Consolidate the way JWK sets are loaded


sberyozkin:
At the moment the JWT sets are dealt with in two different places. First `mp.jwt.publickey.location` is checked if it points to a JWK set resource in `KeyLocationResolver`. Next if an internal `followMpJwt11Rules` is set then it is assumed that instead of  `mp.jwt.publickey.location` `jwksUri` was set and the keys are loaded not in the resolver but in the parser. So we have two ways to point to JWKs, only one of them is used in smallrye-jwt while the `jwksUri` path is used via a `followMpJwt11Rules` diversion from Thorntail.  

This really needs to change and `followMpJwt11Rules` needs to go as it is confusing (and will become more so when we move to newer versions of MP JWT). Instead `KeyLocationResolver` should keep all the code for loading the keys, for example, it can check if the location starts from `https` etc.
`JWTAuthContextInfo` `jwksUri` should also be renamed to `publicKeyLocation` or similar, otherwise at the moment, if is actually used to load PEM keys as well, it is initialized with `mp.jwt.publickey.location` which adds the extra confusion.

sberyozkin:
Good opportunity to fix it now given that we plan to release `2.0.0`

Issue #81, opened 2019-06-20
Change smallrye.jwt.groups.path to smallrye.jwt.path.groups for a more compact and better scoped configuration 


sberyozkin:
This property was introduced in the current SNAPSHOT so changing it now is fine.
`smallrye.jwt.path.groups` will allow for a more compact configuration when multiple paths have to be supported, example, the paths to the custom `groups` and `sub` claims:
```
path:
  groups: a/b
  sub: c/d
```
instead of 

```
groups:
  path: a/b
sub: 
  path: c/d
```

PR #87, opened 2019-06-26
Key resolution optimizations and other minor updates

### LINKED ISSUE #83: Introduce key format and JWK 'kid' properties and optimize PublicKey creation
At the moment 'KeyLocationResolver' loads a public key from the external resource on every request. This can be mostly optimized which will need to be done anyway (have a local volatile PublicKey property should do it for PEM or single JWK files). 

But if it is a JWK set (which may contain N entries) then the problem is that the incoming 'kid' property can vary in theory so we can't settle on the single public key so iterating over the set every time can be expensive. In reality though it would be better and typically expected to have the same 'kid' - which would be managed at the key rotation point ex in the OIDC server. Expecting that every client may have used a different private key to sign is not really realistic from the security POV and as such the floating selection of keys from the JWK set can and should be optionally restricted to the specific 'kid' - and this will let it optimize the PublicKey creation as well.

Furthermore, with a new proiperty for specifying the key store format (PEM file, JKS store (in the future), JWK, JWKSet) the key selection can be further optimized.

It is related to #82 (we would not be able to do the above if the control of the JWK selection would stay inside Jose4J) so I'll fix both issues with a single PR

sberyozkin:
Fixes #83, #82, #81

All the key types are dealt with in `KeyLocationResolver` now so it will be easier to control it further, best attempt at resolving and calculating the public key only once.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
@@ -49,7 +49,7 @@ public class KeyLocationResolver implements VerificationKeyResolver {
     private static final Logger log = Logger.getLogger(KeyLocationResolver.class);
 
     // The 'content' and 'httpsJwks' fields are used to keep the key content and mutually exclusive.
-    // 'content' represents the key(s) loaded from all but the HTTPS URL based resources.
+    // 'content' represents the key(s) loaded from all but the HTTPS URL based JWK set.
     private String content;
     // 'httpsJwks' represents the JWK set loaded from the HTTPS URL.
     private HttpsJwks httpsJwks;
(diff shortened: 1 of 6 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -45,77 +47,113 @@
  */
 public class KeyLocationResolver implements VerificationKeyResolver {
     private static final Logger log = Logger.getLogger(KeyLocationResolver.class);
-    private String location;
     private String content;
+    private HttpsJwks httpsJwks;
```
Ladicek: Would be good to add a comment that `content` and `httpsJwks` are mutually exclusive.

sberyozkin:
@kenfinnigan @Ladicek 

Thanks. Yes, it is not quite ideal because the HTTPs one is actually a cache which refreshes periodically, but at the moment I've just wanted to get all the key related code in one place, and we can improve it if proves possible during the next iterations

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: OK

sberyozkin:
@kenfinnigan or @Ladicek can you please approve again and help me stop keeping updating the comments :-)

sberyozkin:
@kenfinnigan and @Ladicek, going ahead with the merge as you did approve and I've updated the comments, so all should be good

Ladicek:
Sure :-)

PR #90, opened 2019-06-27
Fixed whitelistAlgorithms configuration. Optional of List not mapped correctly.

radcortez:


radcortez:
Hey @sberyozkin how are you? :)

I remember there was a reason why I didn't wrap the whitelistAlgorithms configuration in an Optional. It doesn't work well with Lists. It would map it as a String and then the cast to List will fail, so I had to change it to a String and convert to a List manually.

Issue #84, opened 2019-06-25
Optimize the utility code for setting and calculating the path properties


sberyozkin:
Right now `SmallryeJwtUtils` has a nearly identical code for setting the `sub` and `group` path properties and `DefaultJWTTokenParser` has a nearly identical code for resolving those paths.
In the former case both functions can delegate to the internal one, while in the latter the common function can look for a final segment and let the cast be done in the `sub` and `groups` path code branches

sberyozkin:
@radcortez Are you still OK with doing this minor update ? Np at all if you are busy, I can do it as well, minor effort is required

radcortez:
@sberyozkin sure. if you can wait ill next Monday, I'll do it. (I'll try to make it sooner tomorrow).

sberyozkin:
@radcortez thanks, I've just pinged you because we discussed this refactoring, so I wanted to avoid doing this update if you were already on it. I might as well move some code around in my next PR and I will update you FYI :-)
Cheers

sberyozkin:
@radcortez FYI, https://github.com/smallrye/smallrye-jwt/pull/93/files, I did some tweaks, and minor updates, did not save very much in the parser, but I suspect the casts can be also optimized, may be some other time :-)
thanks

radcortez:
+1 :)

PR #93, opened 2019-06-28
Moving the tck module into testsuite/basic and minor refactorings

### LINKED ISSUE #62: Move the tck module into testsuite/basic similarly to smallrye-rest-client
Right now the tck module tests are not running the real TCK tests against a server, so the proposal is to move the current tests into a new `testsuite/basic` module and at some later stage a testsuite/tck can be added as well (may be the TCK can be run against WildFly and the authentication mechanism Mike Edgar is working upon, etc).

sberyozkin:
Fixes #62, #84


sberyozkin:
Opened #94 to take care of setting up the tck suite run

kenfinnigan:
I'm not sure `testsuite/basic` is enough of an indication that it's a TCK.

What's the need for the move, non TCK testsuites? If so, I'd suggest moving the current ones to `testsuite/tck` instead

sberyozkin:
@kenfinnigan 
My understanding the existing 'tck' module is not really for running MP JWT TCK. This is what [was there originally](https://github.com/smallrye/smallrye-jwt/tree/1.1.0/tck/src), I guess it was a copy of some of the MP JWT TCK code/resources, I'm not sure [this suite is actually run](https://github.com/smallrye/smallrye-jwt/blob/1.1.0/tck/tck-suite.xml) and it is mostly commented out.
The 'tck' module [has grown](https://github.com/smallrye/smallrye-jwt/tree/master/tck/src/test/java/io/smallrye/jwt) since 1.1.0. I found using some of the TCK resources handy, such as the private key, etc, but it is not really TCK.
So my plan was basically move it all into `testsuite/basic` and continue adding tests there, and in time try to get a complete MP JWT TCK run (no idea how it would have to be done, may be with WildFly, since smallrye-jwt now ships the EE security module and the filters)...
Does it sound reasonable ?

radcortez:
It would be interesting if we could have the ability to run the MP TCK from here directly. 

Right now, if you do a change you have no idea if you are breaking the TCK.

sberyozkin:
@radcortez No TCK is run is described above. The goal of this PR is to prepare for having an actual TCK being run

radcortez:
Yeah I understood :) 

Just reinforcing that we should do it :)

kenfinnigan:
The `tck` should most definitely be running the MP JWT TCK, that was its original intent. The TCK is included as a dependency, but it appears that in the `tck-suite.xml` only a single test is being run.

That is not what should be happening. We now have no way to know whether the TCK ever passed.

The TCK should be added and executing ASAP

sberyozkin:
@kenfinnigan I know, this is why I've started refactoring. There is a bunch of tests there which have nothing to do with the actual TCK, so rather than run them alongside with the actual TCK in the same module, I'd like to move them to `testsuite/basic`, so that a clean `testsuite/tck` TCK runner module can be created.
Doe it make sense ?

kenfinnigan:
Sure, my concern was more around that I thought `/tck` was actually running the TCK, but it appears to be mostly disabled

sberyozkin:
@kenfinnigan OK. Yes, I'll have to take care of getting the actual TCK running with  #94, it will have a `tck` module reintroduced.
Can you approve this PR :-) ?

testsuite/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+    <artifactId>smallrye-jwt-parent-1.1</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+  </parent>
+
+  <packaging>pom</packaging>
+  <artifactId>smallrye-jwt-testsuite-parent-1.1</artifactId>
+  <name>SmallRye: MicroProfile JWT - Testsuite Parent</name>
+
+  <modules>
+    <module>basic</module>
```
kenfinnigan: This needs plugin configuration the same as https://github.com/smallrye/smallrye-config/blob/master/testsuite/pom.xml#L38 to prevent any testsuite from being released to Nexus

sberyozkin: @kenfinnigan have a look please, also removed some basic/pom.xml dependencies which are not used

testsuite/basic/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.22.2</version>
```
kenfinnigan: Not critical, but the version of the plugin should be coming from the parent

sberyozkin:
@kenfinnigan thanks, I've updated it to reuse the surefire plugin version from the parent, I can merge once the build has passed

Issue #102, opened 2019-07-12
NullPointerException when claims contain null values


nderwin:
When the value of a claim is an array that contains a null entry, a NullPointerException is thrown when the value is added to the `JsonArrayBuilder` in `JWTCallerPrincipal`.  

It looks like `jsonValue` in `JWTCallerPrincipal.wrapClaimValue()` should be initialized to `JsonValue.NULL` instead of `null`.

PR #103, opened 2019-07-12
Fix for #102

nderwin:
* added a test dependency on the Json implementation
* added a unit test to make sure the changes didn't break existing functionality, also adding test cases for null values
* changed the default returned value to a JsonValue of null, instead of a raw null

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTCallerPrincipal.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTCallerPrincipal.java
@@ -16,8 +16,6 @@
  */
 package io.smallrye.jwt.auth.principal;
 
-import static javax.json.JsonValue.NULL;
-
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -195,7 +193,7 @@ public abstract class JWTCallerPrincipal implements JsonWebToken {
                 if (element instanceof String) {
                     arrayBuilder.add(element.toString());
                 } else if (element == null) {
-                    arrayBuilder.add(NULL);
+                    arrayBuilder.add(JsonValue.NULL);
                 } else {
                     JsonValue jvalue = wrapClaimValue(element);
                     arrayBuilder.add(jvalue);
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTCallerPrincipal.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTCallerPrincipal.java
@@ -155,6 +155,8 @@ public abstract class JWTCallerPrincipal implements JsonWebToken {
                 builder.add(entry.getKey(), flag);
             } else if (entryValue instanceof String) {
                 builder.add(entry.getKey(), entryValue.toString());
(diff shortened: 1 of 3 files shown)


nderwin:
Ooh, test failures I didn't see.  Closing this to fix.

sberyozkin:
@nderwin Thanks for the PR, can you please type a sample JWT (in JSON format), I'd like to see which part of JWTCallerPrincipal throws NPE. The PR looks fine, but I'm not sure if both of your changes are needed or only one, difficult to appreciate if there could be some side-effects. 
Specifically I'm not sure that `JsonValue.NULL` has to be returned as a default value. There well may be a user code which already checks some claim values for `null`
Thanks

nderwin:
Absolutely!  I also share your concern that there may be existing code expecting a Java null instead of a JSON null, so it may be necessary to limit the changes to values inside of collections.  An example JWT would be something like this:

```
{
  "sub": "some@body.com",
  "customClaim": {
    "customValue1": [
      "read"
    ],
    "customValue2": [
      null
    ],
    "customValue3": [
      "admin"
    ]
  },
  "exp": 1562957605
}
```

The trouble I was running into was with `customValue2` above - when the code would attempt to add the null to the JSON array build, it kicked it out with an exception (on purpose, looking at the source code for the implementation).

sberyozkin:
@nderwin Thanks. I've debugged, and the only place in the code which IMHO needs an update is [this one](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTCallerPrincipal.java#L193): add a `JsonValue.NULL`  if it is a null value.

Can you please update your PR ?

nderwin:
Sure!  Just to be clear, we're talking about keeping the change in the `replaceMapClaims` method, and reverting the change in `wrapClaimValue`?

sberyozkin:
@nderwin I believe the only change to address the issue you reported is needed inside the `wrapClaimValue` method at the point where the JsonArray is being prepared.

implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTCallerPrincipal.java
```diff
@@ -192,6 +194,8 @@ protected JsonValue wrapClaimValue(Object value) {
             for (Object element : list) {
                 if (element instanceof String) {
                     arrayBuilder.add(element.toString());
+                } else if (element == null) {
+                    arrayBuilder.add(NULL);
```
sberyozkin: @nderwin thanks, this looks good, only a minor cosmetic issue: NULL is statically imported while just few lines above JsonValue.TRUE/FALSE are used. Lets make it consistent, I don't mind much if the static imports will be used, though have a minor preference for the `JsonValue` qualification in the code as it is easier to see what kind of NULL or Boolean values the code is dealing with, but you choose :-)

nderwin: Ah, IDE auto-imports strike again.

sberyozkin:
@nderwin thank you !

PR #104, opened 2019-07-15
Correct processing of the standard scope claim

### LINKED ISSUE #98: The scope to group claim mapping will not work
The [scope](https://tools.ietf.org/html/rfc6749#section-3.3) claim, if present, will have the value such as `a b c` as opposed to the expectation it would be an array. So if the mapping of `scope` to `groups` is needed then it won't work. A simple check that if it is a `scope` then extract the roles from a string should do

sberyozkin:
Fixes #98


sberyozkin:
@Ladicek thanks, `scope` is a standard OAuth2 claim and is used in OIDC as well, and I realized it was not done correctly in smallrye-jwt after talking to Loic who did a Quarkus OAuth2 PR

PR #107, opened 2019-07-15
Support for the keys on the file system

### LINKED ISSUE #96: Load mp.jwt.verify.publickey.location from current folder
As stated in [MP-JWT Spec 1.1](https://github.com/eclipse/microprofile-jwt-auth/blob/master/spec/src/main/asciidoc/configuration.asciidoc#relative-path), 

> Relative or non-URL paths supplied as the location are resolved in the following order:
>  1. `new File(location)`
>  2. `Thread.currentThread().getContextClassLoader().getResource(location)`

However the first case is not working.

In my `microprofile-config.properties`, I have:
```
mp.jwt.verify.publickey.location=jwk-signature.json
```
If I put `jwk-signature.json` in the current folder and run
```
java -jar target/my-quarkus-runner.jar
```
I get:
```
2019-07-03 11:32:37,006 WARN  [io.sma.jwt.aut.pri.DefaultJWTTokenParser] (executor-thread-1) Token is invalid: JWT processing failed. Additional details: [[17] Unable to process JOSE object (cause: org.jose4j.lang.UnresolvableKeyException: Failed to load a key from: jwk-signature.json): JsonWebSignature{"alg":"RS256","kid":"..."}->...]
```

sberyozkin:
Fixes #96

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
@@ -17,11 +17,14 @@
 package io.smallrye.jwt.auth.principal;
 
 import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringReader;
 import java.io.StringWriter;
+import java.net.URI;
 import java.net.URL;
 import java.security.Key;
 import java.security.PublicKey;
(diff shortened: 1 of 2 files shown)


testsuite/basic/src/test/java/io/smallrye/jwt/KeyLocationResolverTest.java
```diff
@@ -76,13 +76,23 @@ public void testVerifyWithJwkKeyWithNonMatchingKidFromSet() throws Exception {
...
     @Test
-    public void testVerifyWithPemKeyWithMatchingKid() throws Exception {
-        verifyToken("key3", null, "publicKey.pem");
+    public void testVerifyWithClassPathPemKey2() throws Exception {
+        verifyToken("key3", null, "classpath:publicKey.pem");
+    }
+
+    @Test
+    public void testVerifyWithFileSystemPemKey() throws Exception {
+        verifyToken("key3", null, "target/test-classes/publicKey.pem");
```
Ladicek: There's no prefix here, so it should be a classpath resource, but `target/test-classes/publicKey.pem` seems more like a filesystem path?

Ladicek: Aha, that's intentional. All good :-)

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -157,21 +160,36 @@ private PublicKey tryAsJWK(JsonWebSignature jws) throws UnresolvableKeyException
...
+                is = getAsFileSystemResource(location.getRawSchemeSpecificPart());
+            } else {
+                is = new URL(authContextInfo.getPublicKeyLocation()).openStream();
+            }
         } else {
-            URL locationURL = new URL(location);
-            is = locationURL.openStream();
+            is = getAsClasspathResource(authContextInfo.getPublicKeyLocation());
+            if (is == null) {
+                is = getAsFileSystemResource(authContextInfo.getPublicKeyLocation());
```
Ladicek: As rightfully pointed out in #96, the spec says

> Relative or non-URL paths supplied as the location are resolved in the following order:
>  1. `new File(location)`
>  2. `Thread.currentThread().getContextClassLoader().getResource(location)`

But we first attempt option 2, and only then option 1. Should we swap them?

sberyozkin: @Ladicek I thought I'd avoid the extra penalty until the TCK enforces it, as it is highly unlikely that the same named resource would exist both on the classpath and the filesystem, having it on the classpath is more likely. I don't mind much swtching the order though for 2.0.2, will open an issue.

PR #109, opened 2019-07-24
Resolve several issues reported by Sonar

### LINKED ISSUE #101: Review the code issues and address some/all of them
https://sonarcloud.io/project/issues?id=smallrye_smallrye-jwt&resolved=false

MikeEdgar:
Fixes #101 

Aggregate code coverage results from implementation + TCK tests for Sonar reporting.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/JsonUtils.java
+++ b/implementation/src/main/java/io/smallrye/jwt/JsonUtils.java
@@ -1,0 +1,87 @@
+package io.smallrye.jwt;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import javax.json.Json;
+import javax.json.JsonArray;
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
+import javax.json.JsonValue;
+
+public class JsonUtils {
+
+    private JsonUtils() {
+    }
+
+    public static JsonObject replaceMap(Map<String, Object> map) {
+        JsonObjectBuilder builder = Json.createObjectBuilder();
+        for (Map.Entry<String, Object> entry : map.entrySet()) {
+            Object entryValue = entry.getValue();
+            if (entryValue instanceof Map) {
+                @SuppressWarnings("unchecked")
+                JsonObject entryJsonObject = replaceMap((Map<String, Object>) entryValue);
+                builder.add(entry.getKey(), entryJsonObject);
+            } else if (entryValue instanceof List) {
+                JsonArray array = (JsonArray) wrapValue(entryValue);
+                builder.add(entry.getKey(), array);
+            } else if (entryValue instanceof Long || entryValue instanceof Integer) {
+                long lvalue = ((Number) entryValue).longValue();
+                builder.add(entry.getKey(), lvalue);
+            } else if (entryValue instanceof Double || entryValue instanceof Float) {
+                double dvalue = ((Number) entryValue).doubleValue();
+                builder.add(entry.getKey(), dvalue);
+            } else if (entryValue instanceof Boolean) {
+                boolean flag = ((Boolean) entryValue).booleanValue();
+                builder.add(entry.getKey(), flag);
+            } else if (entryValue instanceof String) {
+                builder.add(entry.getKey(), entryValue.toString());
+            }
+        }
+        return builder.build();
+    }
+
+    private static JsonArray toJsonArray(Collection<?> collection) {
+        JsonArrayBuilder arrayBuilder = Json.createArrayBuilder();
+
+        for (Object element : collection) {
+            if (element instanceof String) {
+                arrayBuilder.add(element.toString());
+            } else if (element == null) {
+                arrayBuilder.add(JsonValue.NULL);
+            } else {
+                JsonValue jvalue = wrapValue(element);
+                arrayBuilder.add(jvalue);
+            }
+        }
+
+        return arrayBuilder.build();
+    }
+
+    public static JsonValue wrapValue(Object value) {
+        JsonValue jsonValue = null;
+
+        if (value instanceof JsonValue) {
+            // This may already be a JsonValue
+            jsonValue = (JsonValue) value;
+        } else if (value instanceof String) {
+            jsonValue = Json.createValue(value.toString());
+        } else if ((value instanceof Long) || (value instanceof Integer)) {
+            jsonValue = Json.createValue(((Number) value).longValue());
+        } else if (value instanceof Number) {
+            jsonValue = Json.createValue(((Number) value).doubleValue());
+        } else if (value instanceof Boolean) {
+            jsonValue = (Boolean) value ? JsonValue.TRUE : JsonValue.FALSE;
+        } else if (value instanceof Collection) {
+            jsonValue = toJsonArray((Collection<?>) value);
+        } else if (value instanceof Map) {
+            @SuppressWarnings("unchecked")
+            JsonObject entryJsonObject = replaceMap((Map<String, Object>) value);
+            jsonValue = entryJsonObject;
+        }
+
+        return jsonValue;
+    }
+}
--- a/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
+++ b/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
@@ -71,7 +71,7 @@ public class KeyUtils {
      * 
      * @param pemEncoded - pem string for key
      * @return RSA private key instance
-     * @throws Exception - on failure to decode and create key
+     * @throws GeneralSecurityException - on failure to decode and create key
      */
     public static PrivateKey decodePrivateKey(String pemEncoded) throws GeneralSecurityException {
         pemEncoded = removeBeginEnd(pemEncoded);
--- a/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
+++ b/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
@@ -17,16 +17,15 @@
 package io.smallrye.jwt;
 
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.InputStream;
+import java.io.Reader;
 import java.io.StringReader;
+import java.io.UncheckedIOException;
 import java.math.BigInteger;
+import java.security.GeneralSecurityException;
 import java.security.KeyFactory;
-import java.security.KeyPair;
-import java.security.KeyPairGenerator;
-import java.security.NoSuchAlgorithmException;
-import java.security.PrivateKey;
 import java.security.PublicKey;
-import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.RSAPublicKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Base64;
@@ -34,6 +33,7 @@ import java.util.Base64;
 import javax.json.Json;
 import javax.json.JsonArray;
 import javax.json.JsonObject;
+import javax.json.JsonReader;
 
 /**
  * Utility methods for dealing with decoding public and private keys resources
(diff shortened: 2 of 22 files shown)


implementation/src/main/java/io/smallrye/jwt/JsonUtils.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+import java.util.Map;
+
+import javax.json.Json;
+import javax.json.JsonArray;
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
+import javax.json.JsonValue;
+
```
MikeEdgar: The `JsonUtils` class contains several Json wrapping methods previously duplicated elsewhere in the implementation.

implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
```diff
@@ -17,88 +17,55 @@
...
 import javax.json.Json;
 import javax.json.JsonArray;
 import javax.json.JsonObject;
+import javax.json.JsonReader;
 
 /**
  * Utility methods for dealing with decoding public and private keys resources
  */
 public class KeyUtils {
     private static final String RSA = "RSA";
```
MikeEdgar: Removed unused methods.

implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
```diff
@@ -60,39 +59,52 @@ public String getBearerToken() {
         final String bearerValue;
```
MikeEdgar: Extract token retrieval to separate methods to reduce method complexity (Sonar complexity rating)

implementation/src/main/java/io/smallrye/jwt/auth/cdi/CommonJwtProducer.java
```diff
@@ -106,75 +98,4 @@ public String getName(InjectionPoint ip) {
         return name;
     }
```
MikeEdgar: Functionality moved to `JsonUtils`

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -74,6 +70,20 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
         provider.mpJwtPublicKey = Optional.of(publicKey);
         provider.mpJwtLocation = Optional.of(publicKeyLocation);
         provider.mpJwtIssuer = issuer;
+
```
MikeEdgar: Provide same optional defaults given by CDI/MP Config to avoid use of null checks elsewhere in code

testsuite/basic/pom.xml
```diff
@@ -30,12 +30,15 @@
   <name>SmallRye: MicroProfile JWT Basic Testsuite</name>
 
   <dependencies>
-
```
MikeEdgar: Use compile scope to allow JaCoCo report aggregation to function properly.

kenfinnigan:
@sberyozkin or @starksm64 could you take a look at this?

sberyozkin:
@kenfinnigan @MikeEdgar sure, will do asap, first day back from PTO today

implementation/src/main/java/io/smallrye/jwt/auth/cdi/PrincipalProducer.java
```diff
@@ -60,6 +57,7 @@ public String getName() {
         }
 
         @Override
+        @SuppressWarnings("squid:S1168") // Indicate to Sonar that a null return is acceptable
```
sberyozkin: I'd avoid getting these Sonar specific indicators into the code, if some other system  we only need to address the real code issues, not a problem if Sonar reports some false alarms

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -212,9 +222,7 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
 
         JWTAuthContextInfo contextInfo = new JWTAuthContextInfo();
         // Look to MP-JWT values first
-        if (mpJwtPublicKey.isPresent() && !NONE.equals(mpJwtPublicKey.get())) {
-            decodeMpJwtPublicKey(contextInfo);
-        }
+        decodeMpJwtPublicKey(contextInfo);
```
sberyozkin: Unfortunately this will break Thorntail which offers its own properties as well

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: This will likely break Thorntail

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -238,36 +242,25 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
         SmallryeJwtUtils.setWhitelistAlgorithms(contextInfo, whitelistAlgorithms);
 
         return Optional.of(contextInfo);
     }
 
     protected void decodeMpJwtPublicKey(JWTAuthContextInfo contextInfo) {
+        if (!mpJwtPublicKey.isPresent() || NONE.equals(mpJwtPublicKey.get())) {
+            return;
+        }
+
```
sberyozkin: I think all the optimization in this file will have to be reverted, otherwise Thorntail won't be able to use its own properties

implementation/src/main/java/io/smallrye/jwt/JsonUtils.java
sberyozkin: Please make sure the code from `JWTCallerPrincipal` is moved to this class, not the one from `CommonJwtProducer` (which indeed can be dropped)

MikeEdgar: The version from `JWTCallerPrincipal` is what was used. Note that the only significant difference between the two is that iteration of a `Collection` will add `JsonValue.NULL` if the entry is `null`. Also note that the method of creating Json primitives using a `JsonObjectBuilder` and a "tmp" key was replaced with `Json.createValue` factory calls.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
MikeEdgar: I am not sure that I follow you here. The main change here is to alter the conditions to assume CDI is working properly and that `Optional` objects are always present (which they should be for MP config properties). That includes setting the default values in the `create` method which is used by unit tests. I took a look at the [class having the same name in Thorntail](https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/config/JWTAuthContextInfoProvider.java) and I don't believe there is an impact since that class provides its own producer method which uses this class's getters.

Thoughts?

implementation/src/main/java/io/smallrye/jwt/auth/cdi/PrincipalProducer.java
MikeEdgar: Noted - I'll remove the four occurrences of them.

MikeEdgar: Pushed the removal of Sonar warning suppression.

implementation/src/main/java/io/smallrye/jwt/JsonUtils.java
sberyozkin: Re `JsonValue.NULL` - lets deal with it as part of a separate issue (if there will be a specific regression) - the concerns around replacing `null` everywhere was discussed as part of #103.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: Sorry, should've clarified - you moved the property checks inside the individual methods, and also introduced many default values but unfortunately it will break Thorntail which has the legacy code such as [this one](https://github.com/thorntail/thorntail/blob/master/fractions/microprofile/microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/deployment/auth/config/JWTAuthContextInfoProvider.java#L87).

sberyozkin: Actually, it does look good, I looked too quickly. (by the way I might've deleted some of your comments by mistake, meant to remove one of my ones).

MikeEdgar: The defaults should all be the same applied during construction of `JWTAuthContextInfo` except for `expGracePeriodSecs`, which should be changed to 60 in the `orElse` call. It looks like the condition you linked in Thorntail is redundant, but it appears to me that it would not break. The Thorntail condition and the one in the SmallRye super class are referencing the same `Optional` value and performing the same check.

MikeEdgar: Ah, I didn't see your "update" comment :-)

sberyozkin:
@MikeEdgar Please only revert the replacing of `null` with `JsonValue.NULL` for now (we can review it separately) and I will verify Thorntail against your changes. Thanks

MikeEdgar:
> @MikeEdgar Please only revert the replacing of `null` with `JsonValue.NULL` for now (we can review it separately) and I will verify Thorntail against your changes. Thanks

Sure, will do. Just to be clear, that will revert the change from #103, agreed? It will require a change to `JsonUtilsTests#testWrapClaimValueCollectionWithNull` as well.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: Sorry :-)

sberyozkin:
@MikeEdgar No, that one needs to stay :-) because when the individual array value is `null` then using `JsonValue.NULL` is the way forward. A complete replacement may be eventually completed as well but I'd rather have the spec saying it (but to be honest I'd just just for null in my code - which is what also possibly happening in some other user's code as well)

MikeEdgar:
Looking through the code, the only occurrence of `JsonValue.NULL` is for the #103 fix. Did you see it elsewhere or were you concerned that the code in `JsonUtils` replaces a version with `JsonValue.NULL` and one without it?

sberyozkin:
@MikeEdgar looks like I misinterpreted your earlier note where you mentioned `JsonValue.NULL`. Going to test against Thorntail/Quarkus next

sberyozkin:
@MikeEdgar please restore all of `KeyUtils` methods, Thorntail test utility code depends on it, thanks

MikeEdgar:
> please restore all of `KeyUtils` methods

Done

sberyozkin:
@MikeEdgar thanks.
Can you please fix the JavaDoc issues:
```
 Exit code: 1 - /home/runner/smallrye-jwt/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java:74: error: exception not thrown: java.lang.Exception
[ERROR]      * @throws Exception - on failure to decode and create key
```
etc

and have `JWTCallerPrincipal.wrapClaimValue` restored (delegating to JsonUtils would be fine). have a [Quarkus build failure](https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-jwt/runtime/src/main/java/io/quarkus/smallrye/jwt/runtime/auth/ElytronJwtCallerPrincipal.java#L55). 
Note that code would need to be moved to smallrye-jwt so this build issue is a temp issue, but dealing with that code (wrapping of the unknown claims) needs to be done as part of a different issue

MikeEdgar:
@sberyozkin, updates pushed. Let me know if you would like this branch squashed to a single commit.

sberyozkin:
@MikeEdgar seeing errors like this one in Quarkus:
```
[error]: Build step io.quarkus.arc.deployment.ArcProcessor#validate threw an exception: javax.enterprise.inject.spi.DeploymentException: javax.enterprise.inject.AmbiguousResolutionException: Ambiguous dependencies for type java.util.Optional<javax.json.JsonString> and qualifiers [@Claim(standard = preferred_username)]

```
Not sure yet what may be caused this

implementation/src/main/java/io/smallrye/jwt/auth/cdi/RawClaimTypeProducer.java
```diff
@@ -123,10 +122,10 @@ Double getClaimAsDouble(InjectionPoint ip) {
 
     @Produces
     @Claim("")
-    public Optional getOptionalValue(InjectionPoint ip) {
+    public <T> Optional<T> getOptionalValue(InjectionPoint ip) {
```
MikeEdgar: @sberyozkin - I think this may be the cause of the Quarkus error. I'll revert it.

sberyozkin: Cool, thanks for spotting it

Issue #113, opened 2019-08-07
Claim CDI producer methods missing @RequestScoped


MikeEdgar:
I didn't notice this before because I was directly using `JsonWebToken`s, but the producer methods in `ClaimValueProducer`, `JsonValueProducer`, and `RawClaimTypeProducer` do not specify a scope, which defaults to `@Dependent`. When injecting into `@ApplicationScoped` JAX-RS resources (and also `@Stateless` resources), the claims are not updated per each request.

sberyozkin:
Returning null with the request scoped methods will require a [custom proxy](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/cdi/PrincipalProducer.java#L52), will it work for individual claims ?

MikeEdgar:
I think it will not work for individual claims or those inside of `Optional`.

sberyozkin:
OK, please experiment.

sberyozkin:
Will check myself as well with in Thorntail

MikeEdgar:
I take back what I said about the JSON types. This may only be possible with `ClaimValue`, which I think is fine.

mkouba:
@sberyozkin Yes, only `@Dependent` producer methods are allowed to return `null`. For normal scopes you would get an `IllegalProductException`.

sberyozkin:
@mkouba yes, thanks for the confirmation, I recall having to deal with it when updating PrincipalProducer  :-)

PR #115, opened 2019-08-09
Support ClaimValue dynamic calls to JWT claims, add integration tests

MikeEdgar:
This is where I am going with #113. I added some integration tests with Weld to make sure this actually works with a real CDI provider as well. The management/version for the failsafe plugin could probably be moved up into smallrye/parent sooner or later.

A few things to think about:
1. I think `CommonJwtProducer.wrapClaimValue` is no longer needed - not sure of downstream impact in Thorntail, etc. The updated code will get the claim value directly and only wrap in `Optional` if necessary versus defaulting to wrapping it and unwrapping if necessary.
2. I extended `ClaimValueWrapper`, but I believe the changes could be made directly therein if item 1 is true. I think the wrapper should _always_ delegate to the `JsonWebToken` for value lookup.

Thoughts?

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -77,7 +77,11 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>${version.mokito}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.weld</groupId>
+      <artifactId>weld-junit4</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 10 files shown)


sberyozkin:
@MikeEdgar It does make sense, can you please apply them. they can be reverted if necessary

sberyozkin:
@MikeEdgar the actual issue you are targeting is resolved with this PR, confirmed with the test, but I see an MP TCK test called `PublicKeyAsPEMTest` failing in Thorntail, can you check at your end please as well ? 
Please also resolve the javadoc issues

MikeEdgar:
I fixed the JavaDoc and made the modifications from my first comment. I looked at the TCK test case but it seems unrelated. Is there a publicly-accessible error log for the build failure you saw? If not, can you post the error message/trace here?

sberyozkin:
I'll need to debug, it definitely with this PR but have no clear lead at the moment.

sberyozkin:
@MikeEdgar FYI, [this comparison](113_add_claim_producer_scope) was failing. But, the good news, your last commit got it fixed :-). Going to retest everything now

PR #116, opened 2019-08-09
Update to revised artifact/version scheme

kenfinnigan:


kenfinnigan:
@sberyozkin can you take a look and approve?

sberyozkin:
@kenfinnigan, I was considering if we should do a quick release first with some immediate fixes and get them into Thorntail. @MikeEdgar  resolved a bug in #113 which can be of the immediate interest to Thorntail. What do you think ? @Ladicek,  FYI

Ladicek:
I don't mind either way.

sberyozkin:
@kenfinnigan @Ladicek well, we can that fix in with the updated versioning scheme as well :-), but I'm not sure if it can affect the productiization effort ?

Issue #94, opened 2019-06-28
Create a testsuite/tck module for running the real TCK suite


sberyozkin:

sberyozkin:
Perhaps the JAX-RS authentication filter can be stressed for a start though my initial thought was to use the EE security mechanism.

PR #117, opened 2019-08-13
Add audience verification with several tests

### LINKED ISSUE #110: Add configuration to verify and/or require audience claim
Utilizing `JwtConsumerBuilder.setExpectedAudience(Boolean, String...)`. The logic should align with what is currently done to require/validate the issuer claim.

`smallrye.jwt.verify.audiences` (`Optional<Set<String>>`)
`smallrye.jwt.verify.requireaud` (`Optional<Boolean>`)

Potential configurations if incorporated into specification:
`mp.jwt.verify.audiences`
`mp.jwt.verify.requireaud`

*Edit*: make expected audiences property plural to align with JOSE4J API and OpenLiberty implementation.

MikeEdgar:
Fixes #110 

1. Update DefaultJWTCallerPrincipal to delegate `aud` parsing logic to existing specialized JOSE4J method `JwtClaims#getAudience`.
2. Add TCK test dependencies to the implementation to allow use of token generation/signing utilities in unit tests.
3. Delegate validation of `aud` to the `JwtConsumerBuilder`. Comments? I think taking advantage of the built-in validator is the way to go for this claim. The presence of a `smallrye.jwt.verify.aud` config option implies that the `aud` claim is required. Future custom validation of other claims would need to revisit a different general approach if necessary.


sberyozkin:
Hi @MikeEdgar sorry for delay was a bit overwhelmed today.
It looks nice, will have a closer look tomorrow.
One thing I'd like to suggest, can you consider avoiding bring the tck utility classes into the implementation module and instead have the tests requiring them to `testsuite/basic` so that the tests in the implementation module are only the 'simple' ones not linked to tck in any way.

Does it sound reasonable ?
thanks

MikeEdgar:
Yes - that sounds reasonable. 

For a future PR: Are you opposed with aligning the unit tests in testsuite/basic to the actual class under test? There are a number of tests currently in there that are all focused on the various aspects of `io.smallrye.jwt.auth.principal.DefaultJWTTokenParser.parse(String, JWTAuthContextInfo)` and I think it would be a better approach to move them into the `io.smallrye.jwt.auth.principal` package. Many of them also extend `Arquillian`, which I do not believe is necessary in some or all of those cases.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -252,6 +266,7 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
         contextInfo.setExpGracePeriodSecs(expGracePeriodSecs.orElse(null));
         contextInfo.setJwksRefreshInterval(jwksRefreshInterval.orElse(null));
         SmallryeJwtUtils.setWhitelistAlgorithms(contextInfo, whitelistAlgorithms);
+        contextInfo.setExpectedAudience(expectedAudience.orElse(null));
```
sberyozkin: @MikeEdgar can you please add a get method alongside other getters, for the Thorntail extending this provider to check it if needed

implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTAuthContextInfo.java
```diff
@@ -193,4 +196,12 @@ public String getTokenKeyId() {
     public void setTokenKeyId(String tokenKeyId) {
         this.tokenKeyId = tokenKeyId;
     }
+
+    public Set<String> getExpectedAudience() {
```
sberyozkin: I have a minor preference to drop `Expected`, even though it is technically a better qualifier, but it just happens that `JWTContentInfo` has not followed this naming pattern for other claims (`iss`, etc). But if you prefer to keep `Expected` then I'll be fine with it, it is up to you.

sberyozkin:
@MikeEdgar 
No, not really, it is a work in progress, trying to figure out the correct test module structure. Initially the only test which the project had was a single test in the default package in the (currently defunct) `tck` module, with more tests added there over time, all of them not really related to the tck. So I tried to start cleaning it up, and follow the (archived) smallrye-rest-client idea of having a `testsuite` module, with the `tck` (see #94) and the rest of them.
Please consider moving some of `unit` like tests back to the `implementation` where they'd belong better, but if possible avoid pulling the tck dependencies with them :-), may be just copy some utility code if needed into the `impementation/test` as needed, etc
Thanks

sberyozkin:
@MikeEdgar, seeing this TCK failure:
```
java.lang.AssertionError: audvalue is NOT null, FAIL expected [true] but found [false]
	at org.testng.Assert.fail(Assert.java:94)
	at org.testng.Assert.failNotEquals(Assert.java:496)
	at org.testng.Assert.assertTrue(Assert.java:42)
	at org.eclipse.microprofile.jwt.tck.container.jaxrs.RequiredClaimsTest.verifyAudience(RequiredClaimsTest.java:197)
```
Is validating it at the Jose level has the side-effects ?

MikeEdgar:
@sberyozkin , fixed. It was due to the first change I listed. 
> 1. Update DefaultJWTCallerPrincipal to delegate aud parsing logic to existing specialized JOSE4J method JwtClaims#getAudience.

Are you actively working on #94? Having the TCK available here would definitely help prevent something like this from slipping through. I'm happy to help on that if you are not.

MikeEdgar:
> Please consider moving some of the tests back to the `implementation` where they'd belong better

Good idea - it might make sense to do that with a separate completely non-functional change PR.

implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTAuthContextInfo.java
MikeEdgar: If it's OK, I'd prefer to keep it. Mainly, because it conveys intent at a glance.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
MikeEdgar: Added

implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTAuthContextInfo.java
sberyozkin: OK

sberyozkin:
@MikeEdgar I've been hoping to get to it after the PTO which I returned from it last week, and Thorntail TCK tests is what I check the PRs against. If you think you can afford spending some time on #94 then it would be nice, if yes then lets discuss the options at #94 (going to comment there shortly anyway), thanks

PR #119, opened 2019-08-13
Enable none algorithm

gcollin:
The underlying jwt framework used (joes4j) do support non-signed JWT (with alg=none), so why not enable it as well for smallrye ?
This makes testing much easier

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/SmallryeJwtUtils.java
+++ b/implementation/src/main/java/io/smallrye/jwt/SmallryeJwtUtils.java
@@ -35,7 +35,8 @@ import io.smallrye.jwt.auth.principal.JWTAuthContextInfo;
 public class SmallryeJwtUtils {
     private static final Integer MAX_PATH_DEPTH = 4;
     private static final String COOKIE_HEADER = "Cookie";
-    private static final Set<String> SUPPORTED_ALGORITHMS = new HashSet<>(Arrays.asList(AlgorithmIdentifiers.RSA_USING_SHA256,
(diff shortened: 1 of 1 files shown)


kenfinnigan:
I'm not an expert on security, but my initial concern with enabling it would be the possibility of the algorithm remaining as "none" in production, or that the application isn't tested with an appropriate algorithm prior to production.

Will let @sberyozkin and @starksm64 give their thoughts.

gcollin:
That's true. But you have to add the algorithm to the whitelist anyway, so it's not usable without active action.

sberyozkin:
@gcollin thanks for the PR. but we are still letting the users make a path for the unsigned tokens which only really will start making sense when we will support the `JWE` enctyption, when it will be fine if the wrapped token is not signed.
We had similar discussions with the users about making the expiry check optional and probably other properties (even restricting the whitelist not to support the symmetric encryption as it raises all sort of key distribution issues), but our plan is to have no even accidental path into loosening the generic level validation.

What issues are you seeing with the testing ? You can add a tck test support dependency and then just do [this](https://github.com/smallrye/smallrye-jwt/blob/master/testsuite/basic/src/test/java/io/smallrye/jwt/TestTokenRequireSub.java#L24), and you have a signed token. I agree it is a bit involved to set it up, but you do it once and then just reuse the same boilerplate code everywhere :-).
Thanks.

gcollin:
It was more for testing the angular client without the hassle of having an Id provider online.
Nevermind, I will find another way.
Thanks for having a look at it!

sberyozkin:
@gcollin OK, hope you will get it sorted out easily enough, thanks

Issue #85, opened 2019-06-25
Support for KeyStores for keeping the keys


sberyozkin:
Using keystores is quite likely the most used option in practice

sberyozkin:
As part of this issue I will introduce a key type property which will allow to optimize the key loading further

darranl:
+1 to adding the key type property.  FYI we do have some code within the WildFly Elytron project for detecting the type of a keystore, if there was interest we could move this over to a SmallRye project to make it easier to re-use.

A couple of additional points that may be worth considering.
- Obviously the store itself needs a password to open it, but each entry could also have a different password so if multiple entries are to be used this may need to be represented in the config.
- Once the type can be specified sometimes it is desirable to be able to specify the name of the provider to use although that may be more advanced than required here.

A second piece of code we have in the Elytron project that could likely be moved if there was interest is a KeyStore which can wrap another KeyStore to filter the results - similar to what you would get from a database view of a much larger table.

The reason I mention the latter is there may be a point where users would use say a PKCS#11 store but want to filter the keys to a specific set of aliases.

sberyozkin:
@darranl Hi Darran, thanks for the comments, sorry for as delay,
 
> +1 to adding the key type property

I think it may be better for me to create a separate light-weight PR after all, introducing this property only with only few minor code touches, so that you can review and then we can proceed next.

> FYI we do have some code within the WildFly Elytron project for detecting the type of a keystore, if there was interest we could move this over to a SmallRye project to make it easier to re-use

Sounds interesting. Lets have an initial support first then double check, not sure yet if expecting the users to set the key store type could be enough or not.

> Obviously the store itself needs a password to open it, but each entry could also have a different password so if multiple entries are to be used this may need to be represented in the config

Can we say that access to the public keys only requires an alias ? That way it would work the same way it works for a JWK set, the token `kid` would be used to select a given public key, while the `smallrye.jwt.token.kid` property can be used to select a specific public key only.

> Once the type can be specified sometimes it is desirable to be able to specify the name of the provider to use although that may be more advanced than required here

We can definitely support a little bit later, good idea.

> A second piece of code we have in the Elytron project that could likely be moved if there was interest is a KeyStore which can wrap another KeyStore to filter the results - similar to what you would get from a database view of a much larger table

I'm not sure yet how that would be used, but lets keep this option on the table for sure, we can likely get it in once the basic support is in place, I'll be asking you to review :-)

Thanks

darranl:
[...]

PR #120, opened 2019-08-14
fixed potential problems with multithreading

ys-kalyakin:


implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+        KeyLocationResolver keyLocationResolver = new KeyLocationResolver(new JWTAuthContextInfo("wrong_location.pem", null));
+        keyLocationResolver.resolveKey(jsonWebSignature, emptyList());
+    }
+
+    @Test
+    public void testHttpsKeyLocation() throws Exception {
+        JWTAuthContextInfo contextInfo = new JWTAuthContextInfo("https://github.com/my_key.pem", "issuer");
+        contextInfo.setJwksRefreshInterval(10);
+        Headers headers = Mockito.mock(Headers.class);
```
Ladicek: Using `@Mock` for the previous test and programmatic `mock()` for this one looks strange.

implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
```diff
@@ -0,0 +1,72 @@
...
+        Mockito.when(headers.getStringHeaderValue(JsonWebKey.KEY_ID_PARAMETER)).thenReturn("1");
+        JsonWebSignature signature = Mockito.mock(JsonWebSignature.class);
+        Mockito.when(signature.getHeaders()).thenReturn(headers);
+        PublicKey key = Mockito.mock(PublicKey.class);
+        KeyLocationResolver keyLocationResolver = new KeyLocationResolver(contextInfo);
+        keyLocationResolver = Mockito.spy(keyLocationResolver);
+        Mockito.doReturn(key).when(keyLocationResolver).getHttpJwk(Mockito.any(), Mockito.any());
+        assertEquals(key, keyLocationResolver.resolveKey(signature, emptyList()));
+        assertTrue(
+                new FieldReader(keyLocationResolver, KeyLocationResolver.class.getDeclaredField("verificationKey")).isNull());
```
Ladicek: Looking at the package name, `FieldReader` doesn't seem to be considered as public API by Mockito.

Ladicek:
I only looked briefly, but this looks good to me. The change in `KeyLocationResolver.resolveKey` is too big to easily review in GitHub, so someone should take a look at that in more detail :-)

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -208,7 +208,10 @@ private Object findClaimValue(String claimPath, Map<String, Object> claimsMap, S
 
     protected VerificationKeyResolver getKeyResolver(JWTAuthContextInfo authContextInfo) throws UnresolvableKeyException {
         if (keyResolver == null) {
-            keyResolver = new KeyLocationResolver(authContextInfo);
+            synchronized (this) {
```
sberyozkin: But it is `volatile`, the very worst thing that can happen it can be created more than once but ultimately there will be only a single instance used

sberyozkin: Never mind, guess this pattern is better anyway

implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
MikeEdgar: I suggest making `verificationKey` package private so that it can be accessed by unit tests without reflection.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,50 +53,56 @@
...
     // only once and used for all the token verification requests
     private volatile PublicKey verificationKey;
 
-    private JWTAuthContextInfo authContextInfo;
-
     public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws UnresolvableKeyException {
         this.authContextInfo = authContextInfo;
         try {
-            loadContents();
+            AbstractMap.SimpleImmutableEntry<HttpsJwks, String> contents = loadContents();
```
sberyozkin: I'm not sure we need this kind of clean up, The key and value type are never used at the same time so it looks a bit unnatural to me, even though, anti-pattern wise, this style is technically cleaner.

MikeEdgar: Would it be cleaner to have `loadContents` continue to set the instance variables? I agree it looks unnatural to use `Map.Entry` as a way to return two values that are not actually a key/value pair.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,50 +53,56 @@
...
+        if (httpsJwks != null && authContextInfo.getTokenKeyId() == null)
+            return tryAsJWK(jws);
+
+        if (verificationKey == null) {
+            synchronized (this) {
+                if (verificationKey == null) {
+                    PublicKey localKey = tryAsJWK(jws);
+
+                    if (localKey == null) {
+                        try {
```
sberyozkin: Lets keep this code inside `tryAsPEM` but avoid the `verificationKey` assignment there

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,50 +53,56 @@
...
-
-    private PublicKey tryAsPEM() {
-        try {
-            verificationKey = KeyUtils.decodePublicKey(content);
-        } catch (Exception e) {
-            log.debug("Failed to read location as PEM", e);
+        // httpsJwks may refresh itself even if the jwksRefreshInterval is set to 0 as it checks HTTPS Cache-Control header
+        // so the public key will have to be created per request to ensure the rotation (if any) is effective.
+        if (httpsJwks != null && authContextInfo.getTokenKeyId() == null)
+            return tryAsJWK(jws);
```
sberyozkin: I think in general this is better cleanup, the initial attempt to consolidate the key processing inside the resolver was not perfect.  But I'd like to avoid calling `tryAsJwk` twice here since there will again some redundant checks done there. 
I propose to remove `tryAsJwk` completely, but move its code extracting the `kid` into into a `getKid(JsonWebSignature)` method (I will need it anyway for #85 as well to match the kid against the key alias).
Next here a call to tryAsHttpsJwk(jws) will be made, which will get the kid from `getKid(Jws)`

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,50 +53,56 @@
...
+
+        if (verificationKey == null) {
+            synchronized (this) {
+                if (verificationKey == null) {
+                    PublicKey localKey = tryAsJWK(jws);
```
sberyozkin: Please update it to `PublicKey localKey = tryAsJWKContent(jws)` similarly to the HTTPS JWK check

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -104,7 +111,7 @@ private PublicKey tryAsJWK(JsonWebSignature jws) throws UnresolvableKeyException
         String kid = jws.getHeaders().getStringHeaderValue(JsonWebKey.KEY_ID_PARAMETER);
```
sberyozkin: Please move the `kid` extraction code into its own function and remove `tryAsJWK` as suggested above

sberyozkin: Please move the kid extraction code into its own function and remove tryAsJWK as suggested above

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -181,13 +182,13 @@ PublicKey getContentJwk(String content, String kid) throws JoseException {
         return null;
     }
 
-    private void loadContents() throws IOException {
+    private AbstractMap.SimpleImmutableEntry<HttpsJwks, String> loadContents() throws IOException {
```
sberyozkin: Lets keep the original code here to keep things simpler

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: @MikeEdgar agreed

implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
ys-kalyakin: ok

implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
ys-kalyakin: i made `verificationKey` package private

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
ys-kalyakin: ok

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
ys-kalyakin: I did a little different

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
  */
 public class KeyLocationResolver implements VerificationKeyResolver {
-    private static final Logger log = Logger.getLogger(KeyLocationResolver.class);
+    private static final Logger LOGGER = Logger.getLogger(KeyLocationResolver.class);
+    private static final String FAILED_TO_READ_LOCATION_MESSAGE = "Failed to read location as any of JWK, JWKS, PEM. ";
 
     // The 'content' and 'httpsJwks' fields are used to keep the key content and mutually exclusive.
     // 'content' represents the key(s) loaded from all resources but the HTTPS URL based JWK set.
-    private String content;
+    private final Supplier<String> contentSupplier;
```
sberyozkin: Please, just give us back the plain `String` and `HttpsJwk` references, and the original loadContent(), these fancy wrappers make me nervous because I don't what they are to do with the issue this PR is trying to address :-), And those two suppliers add a lot of code too.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
-        }
-        PublicKey key = tryAsJWK(jws);
-        if (key == null) {
-            key = tryAsPEM();
-        }
-        if (key == null) {
-            throw new UnresolvableKeyException("Failed to read location as any of JWK, JWKS, PEM. "
-                    + authContextInfo.getPublicKeyLocation());
+
+        if (jws == null)
```
sberyozkin: jws will be never be null, remove this check please, Jose4J must guarantee it

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
+            throw new UnresolvableKeyException(FAILED_TO_READ_LOCATION_MESSAGE);
+
+        // httpsJwks may refresh itself even if the jwksRefreshInterval is set to 0 as it checks HTTPS Cache-Control header
+        // so the public key will have to be created per request to ensure the rotation (if any) is effective.
+        String kid = getKid(jws);
```
sberyozkin: Instead please pass only `jws` to both `getHttpsJwk` and `getLocalJwk` and use `getKid` from those 2 methods because at this level the `getKid` call is irrelevant for the `tryAsPem` code branch

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
+        HttpsJwks httpsJwks = httpsJwksSupplier.get();
+        if (httpsJwks != null && authContextInfo.getTokenKeyId() == null)
+            return getHttpsJwk(httpsJwks, kid);
+
+        if (verificationKey == null) {
+            synchronized (this) {
+                if (verificationKey == null) {
+                    String content = contentSupplier.get();
```
sberyozkin: As I said just have a plain null check and that is it, the original loadContents takes care or should if it does not to throw the exception if content is null && httpsJwk is null

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
+                        return PublicJsonWebKey.class.cast(currentJwk).getPublicKey();
+                    }
+                }
+            } else if (keys.size() == 1) {
+                return PublicJsonWebKey.class.cast(keys.get(0)).getPublicKey();
             }
-        } else {
-            kid = authContextInfo.getTokenKeyId();
+        } catch (Exception e) {
+            LOGGER.debug("Failed to read location as JWK(S)", e);
```
sberyozkin: Given we are creating the dedicated functions for different locations, please update to ""Failed to read JWK set from the HTTPS location", something like that.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
         }
+        return null;
+    }
 
-        PublicKey publicKey = null;
-
+    PublicKey tryAsLocalJwk(String content, String kid) {
         try {
-            log.debugf("Trying location as JWK(S)...");
+            JsonObject jwk = null;
```
sberyozkin: Can we call it 'jsonJwk' or 'jwkJson' ?

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
         }
 
-        return publicKey;
+        return null;
     }
 
-    PublicKey getHttpJwk(HttpsJwks httpsJwks, String kid) throws JoseException, IOException {
-        List<JsonWebKey> keys = httpsJwks.getJsonWebKeys();
+    private String getKid(JsonWebSignature jws) throws UnresolvableKeyException {
+        if (jws.getHeaders() == null)
```
sberyozkin: the JWS without the headers is invalid and Jose4J will fail earlier if it were the case, please remove this check

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -52,185 +52,226 @@
...
-                for (int i = 0; i < keys.size(); i++) {
-                    JsonObject currentJwk = keys.getJsonObject(i);
-                    if (kid.equals(currentJwk.getString(JsonWebKey.KEY_ID_PARAMETER, null))) {
-                        jwk = currentJwk;
-                        break;
-                    }
-                }
-            } else if (keys.size() == 1) {
-                jwk = keys.getJsonObject(0);
+        @Override
```
sberyozkin: Please, as I said, let me see the original loadContent with some simple cleanup if needed and I will be happy :-)

sberyozkin:
Hi @ys-kalyakin OK, I have left quite a few comments, the PR is targeting an important issue, thanks for finding the time to create the PR, but lets stay focused on the issue of the PR, it is nearly there :-)

Cheers

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
ys-kalyakin: ok

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
ys-kalyakin: ok

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
ys-kalyakin: loadContent is dangerous, it can be called not only from the constructor and it changes state of the object

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -77,110 +76,108 @@ public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws Unresolvab
...
-    }
-
-    private PublicKey tryAsPEM() {
-        try {
-            verificationKey = KeyUtils.decodePublicKey(content);
-        } catch (Exception e) {
-            log.debug("Failed to read location as PEM", e);
+        // httpsJwks may refresh itself even if the jwksRefreshInterval is set to 0 as it checks HTTPS Cache-Control header
+        // so the public key will have to be created per request to ensure the rotation (if any) is effective.
+        if (httpsJwks != null && authContextInfo.getTokenKeyId() == null)
```
sberyozkin: Can you please wrap it with the curly parentheses, I believe it is the pattern we use even for the single statement ifs. (and please push the comment above inside to make that code inside the curly brackets look bigger :-) and the comment will be better scoped)

sberyozkin:
@ys-kalyakin Looks much nicer IMHO, thank you. I've left a minor comment only, please do a quick update and it is done as far as I'm concerned at least, unless the guys have more comments. In meantime I'll be testing this PR against Thorntail and Quarkus. Cheers

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -77,110 +76,108 @@ public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws Unresolvab
...
+            return getHttpsJwk(httpsJwks, jws);
```
sberyozkin: Actually, couple of more minor comments - we don't need now to pass `httpsJwks`

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -77,110 +76,108 @@ public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws Unresolvab
...
+
+        if (verificationKey == null) {
+            synchronized (this) {
+                if (verificationKey == null) {
+                    PublicKey localKey = tryAsLocalJwk(content, jws);
```
sberyozkin: No need to pass the `content` as a parameter

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -77,110 +76,108 @@ public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws Unresolvab
...
+
+                    if (localKey == null) {
+                        localKey = tryAsPEM(content);
```
sberyozkin: Same here about the `content`

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -77,110 +76,108 @@ public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws Unresolvab
...
+                    }
+                    if (localKey == null) {
+                        throw new UnresolvableKeyException(FAILED_TO_READ_LOCATION_MESSAGE
+                                + authContextInfo.getPublicKeyLocation());
+                    }
+                    verificationKey = localKey;
```
sberyozkin: Oh well, after looking at this code for a while, I've realized the new code has changed the original behaviour and introduced a bug. Basically, if we work with the local JWK sets we can not assign the 'localKey' to the `verificationKey` unless this (old) [condition](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java#L128) is true: this is because JWK set may have many JWKs, each of them with its own `kid`, so unless that condition is true, and more specifically, `authContextInfo.getTokenKeyId() != null` is true (because if this property is set it means the users wants to deal with a specific `kid` only and hence the fact JWK set has many keys does not matter), we can not initialize `verificationKey`.
There should've been a test, but well, I've spotted it now.
I honestly don't know now what are we doing with this KeyLocationResolver code.
`verificationKey` can not be not null in 2 cases:
1. tryAsPEM - PEM key file has no KID to match against
2. JWK set and authContextInfo.getTokenId() != null
And it can't be both 1 and 2 for a given KeyLocationResolver instance.

So why can't we return to the drawing board again and see what are we trying to protect against, as far as the thread safety is concerned, in `KeyLocationResolver` ? Explain me please, why just having `volatile` is not enough given what I have just explained.

I think `getKid(jws)` is definitely the good thing that came out of our discussions.
The other thing which I've realized is that tryAsPEM can be done from the constructor immediately after loadContent().

sberyozkin:
@ys-kalyakin I appreciate it can be frustrating since we have done few rounds of `KeyLocationResolver` refactoring in this PR but let me suggest simply return to the original code, and 
1) add the minor getKid(jws) refactoring we have come up with here (it will affect the original getKeyAsJWK() only) - I will 100% call this method from the JKS store code branch later
2) Keep your capital LOGGER change :-)

3) Note there is absolutely no risk that the `verificationKey` can be assigned to different value public keys by different threads, because it is either PEM only or JWK only when the verification key can be assigned, but I guess we can have some probability of the redundant instance creation, on either PEM only or JWK only paths. I think in the very worst case we will have an only once GC work to have those extra instance dropped because the volatile qualifier will ensure the window for creating the extra instances is very small. So I feel having 'volatile` should be enough.

However if you do prefer to follow the proper volatile assignment pattern for the `verificationKey` and you do not see a clean option to do with the original code as part of point 3) then please do so, but I'd like to encourage you to start again from the original code, and please keep in mind what I have said earlier, I'd not like the local JWK code path where no verification key is assigned be synchronized.

Thank you and look forward to merging this PR sooner rather than later :-)

sberyozkin:
@ys-kalyakin thank you :-), glad you agreed to cancel most of the KeyLocationResolver changes for now.
Lets do this: I will verify the PR against Thorntail/Quarkus on Mon, have to sign off for the weekend.
Once I merge, please have another look at KeyLocationResolver and if you find a clean way to introduce a perfect `verificationKey`, now that you know more about some 'subtleties' (caused by the fact the resolver has to deal with many key types), then the next PR focusing on KeyLocationResolver only will be welcome.
In any case, thanks for the contribution, enjoy the weekend and please get more PRs in time
Cheers

PR #122, opened 2019-08-20
Support for custom group claims with the comma-separated values

### LINKED ISSUE #121: Make mapping from a single string to the groups claim more generic
At the moment, if a path to the custom claim containing the group values points to a non-array, single string, then it is only processed if it is a well known `scope` claim which has the space separated values.
The Quarkus discussion around the JHipster integration highlighted a limitation of the above approach, since the custom `auth` claim created by JHipster has the [comma-separated values](https://github.com/jhipster/jhipster-sample-app/blob/master/src/main/java/io/github/jhipster/sample/security/jwt/TokenProvider.java#L64).
A new `smallrye.jwt.groups-separator` needs to be introduced with the default value " " for the current `scope` support to stay as currently is for the users.

sberyozkin:
Fixes #121

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
@@ -45,7 +45,6 @@ import org.jose4j.lang.UnresolvableKeyException;
 public class DefaultJWTTokenParser {
     private static final Logger LOGGER = Logger.getLogger(DefaultJWTTokenParser.class);
     private static final String ROLE_MAPPINGS = "roleMappings";
-    private static final String SCOPE_CLAIM = "scope";
     private volatile VerificationKeyResolver keyResolver;
 
     public JwtContext parse(final String token, final JWTAuthContextInfo authContextInfo) throws ParseException {
@@ -164,13 +163,18 @@ public class DefaultJWTTokenParser {
             if (claimValue instanceof List) {
                 @SuppressWarnings("unchecked")
                 List<String> groups = List.class.cast(claimValue);
-                return groups;
-            } else if (pathSegments.length == 1 && SCOPE_CLAIM.equals(pathSegments[0])) {
-                if (claimValue instanceof String) {
-                    return Arrays.asList(((String) claimValue).split(" "));
+                // Force a check that a list contains the string values only
+                try {
+                    return Arrays.asList(groups.toArray(new String[] {}));
+                } catch (ArrayStoreException ex) {
+                    LOGGER.warnf("Claim value at the path %s is not an array of strings",
+                            authContextInfo.getGroupsPath());
                 }
+            } else if (claimValue instanceof String) {
+                return Arrays.asList(((String) claimValue).split(authContextInfo.getGroupsSeparator()));
             } else {
-                LOGGER.warnf("Claim value at the path %s is not an array", authContextInfo.getGroupsPath());
+                LOGGER.warnf("Claim value at the path %s is neither an array of strings nor string",
+                        authContextInfo.getGroupsPath());
             }
         }
         if (authContextInfo.getDefaultGroupsClaim() != null) {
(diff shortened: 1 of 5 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -162,15 +161,18 @@ private String findSubject(JWTAuthContextInfo authContextInfo, JwtClaims claimsS
...
             if (claimValue instanceof List) {
-                @SuppressWarnings("unchecked")
-                List<String> groups = List.class.cast(claimValue);
-                return groups;
-            } else if (pathSegments.length == 1 && SCOPE_CLAIM.equals(pathSegments[0])) {
-                if (claimValue instanceof String) {
-                    return Arrays.asList(((String) claimValue).split(" "));
+                try {
+                    @SuppressWarnings("unchecked")
+                    List<String> groups = List.class.cast(claimValue);
```
Ladicek: This doesn't make sense. We do `if (claimValue instanceof List) { ... }`, so inside that `if`, we can safely do `List<String> groups = (List<String>) claimValue;`

Ladicek: Well, not _entirely_ safely, because erasure, but you get the idea :-)

Ladicek:
Otherwise LGTM.

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
sberyozkin: @Ladicek, it is not guaranteed to be a String though, or may be I'm missing something :-) ?

Ladicek: `List.class.cast(claimValue);` doesn't guarantee that either :-)

Ladicek: I mean, what I'm trying to say is: `List.class.cast(claimValue);` is IMHO not idiomatic, and we should use `(List<String>) claimValue`.

sberyozkin: @Ladicek I'm so slow, finally got it :-), let me tweak it

sberyozkin: @Ladicek I thought you were concerned about the whole exception catch thing :-)

sberyozkin: @Ladicek done

Ladicek: Ah, yea, catching the exception doesn't make sense either :-) The only way to verify that the list really contains strings and nothing but strings is to manually verify each element.

Plus, `List<String> groups = (List<String>)List.class.cast(claimValue);` is worse then the previous way, you're casting twice :-)

Ladicek: I probably shouldn't have started complaining about this, because it was already there :-)

sberyozkin: @Ladicek  That was me doing it in a rash, forgot to drop the 1st cast.
Actually, just noticed it was changed [here](https://github.com/smallrye/smallrye-jwt/commit/6a9ed997232319c043bbb3ccb7adfe2d6dde5af6#diff-5f4020e440d6d2dd7b7ff27fbe83dd25), originally it was done with a plain List Of String cast, so I believe Sonar must've been complaining about it. Anyway, I'm happy just to use the original List Of String cast if you think it is a better pattern.
But yeah, the cast exception, I was hoping the compiler would be somehow smart enough to translate that into the individual checks I guess. Have to find a clean way to get this exception thrown, hate to iterate manually :-). Have something in mind ? I guess we just need to create a new List and feed the original List into it somehow

Ladicek: I'm totally not a fan of what we're doing with Sonar. Anyway, please feel free to remove the last commit, restore `List.class.cast(...)` and let's be done with it :-)

PR #125, opened 2019-08-22
Optimize KeyLocationResolver, support for PEM certs

### LINKED ISSUE #112: Support X509 certificates in the PEM format
The certificates such as [these ones](https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com) should be supported

### LINKED ISSUE #124: Optimize KeyLocationResolver
`KeyLocationResolver` needs more work to clean it up, and remove the `volatile` qualifier from the `verificationKey`

sberyozkin:
`KeyLocationResolver` has been refactored to have the `verificationKey` calculated, if possible, only once in the constructor. The local JWK set is now preloaded (JSON-parsed) in any case, so if it is has to be checked per-request then only the in-memory representations are checked. `KeyUtils` has been fixed to correctly replace the PEM begin/end lines and a method for decoding the PEM certs has been added.
Fixes #124.
Fixes #112.


implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
```diff
@@ -137,25 +138,55 @@ public static PublicKey decodeJWKSPublicKey(String jwksValue) throws GeneralSecu
...
+     * 
+     * @param pem encoded string with option header/footer
+     * @return a single base64 encoded pem string
+     */
+    private static String removeCertBeginEnd(String pem) {
+        pem = pem.replaceAll("-----BEGIN(.*?)CERTIFICATE-----", "");
+        pem = pem.replaceAll("-----END(.*?)CERTIFICATE-----", "");
         pem = pem.replaceAll("\r\n", "");
         pem = pem.replaceAll("\n", "");
+        pem = pem.replaceAll("\\\\n", "");
```
Ladicek: What does this do? Can the string ever contain literal `\n`? (Not the newline character, but backslash followed by "n".)

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -56,146 +58,124 @@
     private static final String CLASSPATH_SCHEME = "classpath:";
     private static final String FILE_SCHEME = "file:";
 
-    // The 'content' and 'httpsJwks' fields are used to keep the key content and mutually exclusive.
-    // 'content' represents the key(s) loaded from all resources but the HTTPS URL based JWK set.
-    private String content;
+    // The verification key can be calculated only once and used for all the token verification requests.
+    // It will be created in the constructor if the PEM or the local JWK(S) content is available.
+    // 'smallrye.jwt.token.kid' has to be set for the verificationKey to be created from the local JWK(S). 
+    PublicKey verificationKey;
```
Ladicek: Why not `private`?

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -56,146 +58,124 @@
...
 
+        String content = readKeyContent(keyLocation);
+
+        // Try to init the verification key from the local PEM or JWK(S) content
+        verificationKey = tryAsPEMPublicKey(content);
+        if (verificationKey == null) {
+            verificationKey = tryAsPEMCertificate(content);
+        }
+        if (verificationKey == null) {
+            localJwks = loadLocalJwks(content);
```
Ladicek: This should guard against the content not being valid JWK / JWKS. The `loadLocalJwks` and `getLocalJwk` methods will probably throw in such case, and we should catch here and do some debug logging.

And we should perhaps check in this class constructor if `verificationKey` is still `null` after `initializeKeyContent` returns. Currently, the constructor expects `initializeKeyContent` to either set the `verificationKey` field, or throw. Some methods, however, don't throw, while some do. This is very hard to follow.

Ladicek: Alternatively, we should let all methods just throw.

Ladicek: Actually my comment above isn't entirely accurate; I'm aware of that. This entire class is very very hard to understand.

Ladicek: I'm gonna say one thing: the common wisdom is that many small functions are better than one single big function. The common wisdom is a piece of crap. The functions are scattered around with no order, their names are mildly confusing, and the control flow is far from obvious.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -213,21 +193,80 @@ private void loadContents() throws IOException {
...
         }
-        content = contents.toString();
+        return null;
+    }
+
+    private static JsonWebKey createJsonWebKey(JsonObject jsonObject) throws Exception {
+        return JsonWebKey.Factory.newJwk(JsonUtil.parseJson(jsonObject.toString()));
     }
 
     private static InputStream getAsFileSystemResource(String publicKeyLocation) throws IOException {
```
Ladicek: Body of this method should be:

```java
        try {
            return new FileInputStream(publicKeyLocation);
        } catch (FileNotFoundException e) {
            return null;
        }
```

This `return f.exists() ? new FileInputStream(f) : null;` just creates a false sense of security, but doesn't really prevent throwing.

Ladicek: Or at the very least, wrap the `return f.exists() ? new FileInputStream(f) : null;` into a `try` block, if you want to minimize number of thrown exceptions.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
Ladicek: Aha, because tests.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -213,21 +193,80 @@ private void loadContents() throws IOException {
         }
 
         if (is == null) {
-            throw new IOException("No resource with the named " + authContextInfo.getPublicKeyLocation() + " location exists");
+            throw new IOException("No resource with the named " + keyLocation + " location exists");
```
Ladicek: I'd suggest `throw new IOException("No resource exists at " + keyLocation);`.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
Ladicek: OK, I went through the class in an IDE and it looks mostly fine, but I'm not sure about `loadLocalJwks`. I think it should [attempt to] never throw, and instead return `null`. That's what all methods in this class do.

implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
sberyozkin: Yes, had to add it to deal with [such certs](https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com), one of them has been copied to the test resources.

Ladicek: Ah, very interesting, thanks!

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: @Ladicek, thanks for giving it a thorough check :-). Yes, I was conscious as well about various methods mixed in, so at least I tried to push all the static methods at the end of the class. My main target was to optimize the creation of the keys.  Pushed some code into the standalone functions like reading the input stream, to simplify the testing.
Going to make sure a `null` is returned, indeed it would be more consistent

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: Sure will do, a user provided this message so I thought of keeping it as it was

sberyozkin:
@Ladicek, thanks for the quality review. I'm sure there will be more improvements to this code coming further down the line. Right now I'm happy a number of duplicate and confusing branches have gone away and the key calculation is more efficient :-)

Ladicek:
Thanks! I agree :-)

sberyozkin:
@Ladicek, can you please check again, thanks

sberyozkin:
@Ladicek, np at all :-), happy to keep cleaning up my code :-)

Issue #135, opened 2019-10-10
Public key from http (without SSL) on Windows


wicksim:
We're facing a problem in our local environment when trying to get the public key from a non-SSL-protected server:

> java.nio.file.InvalidPathException: Illegal char <:> at index 4: http://localhost:30000/auth/realms/test/protocol/openid-connect/certs
> 	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
> 	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
> 	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
> 	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
> 	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
> 	at sun.nio.fs.AbstractPath.resolve(AbstractPath.java:53)
> 	at io.quarkus.runner.RuntimeClassLoader.findApplicationResourceContent(RuntimeClassLoader.java:446)
> 	at io.quarkus.runner.RuntimeClassLoader.getResourceAsStream(RuntimeClassLoader.java:149)
> 	at io.smallrye.jwt.auth.principal.KeyLocationResolver.getAsClasspathResource(KeyLocationResolver.java:287)
> 	at io.smallrye.jwt.auth.principal.KeyLocationResolver.readKeyContent(KeyLocationResolver.java:188)
> 	at io.smallrye.jwt.auth.principal.KeyLocationResolver.initializeKeyContent(KeyLocationResolver.java:161)
> 	at io.smallrye.jwt.auth.principal.KeyLocationResolver.<init>(KeyLocationResolver.java:77)
> 	at io.smallrye.jwt.auth.principal.DefaultJWTTokenParser.getKeyResolver(DefaultJWTTokenParser.java:229)
> 	at io.smallrye.jwt.auth.principal.DefaultJWTTokenParser.parse(DefaultJWTTokenParser.java:74)
> 	at io.quarkus.smallrye.jwt.runtime.auth.MpJwtValidator.validateClaimsSet(MpJwtValidator.java:41)
> 	at io.quarkus.smallrye.jwt.runtime.auth.MpJwtValidator.validate(MpJwtValidator.java:35)
> 	at io.quarkus.smallrye.jwt.runtime.auth.MpJwtValidator_ClientProxy.validate(MpJwtValidator_ClientProxy.zig:193)
[...]

PR #138, opened 2019-10-21
Fix/135

wicksim:
Fix for #135


sberyozkin:
@wicksim Thanks for the initial contribution, this will require some more work as even JWKS may be available at HTTP, I'm traveling the next few days, I've already started working on this code as well as part of https://github.com/quarkusio/quarkus/issues/4493 (only did a single key jwkset withhout a kid fix - next is to support HTTP - for all keys and HTTPS - for all keys, not only JWKs).

wicksim:
@sberyozkin OK, so I'll close this one and wait for your changes...

PR #142, opened 2019-11-05
Move NullJsonWebToken into an independent class and add ProviderInjectionSupport

### LINKED ISSUE #132: Improve the CDI code
Right now Thorntail and Quarkus have some CDI code which ideally should be located in smallrye-jwt in the form of `CDIUtils` and may be `AbstractExtension` referring to these `CDIUtils` for Thorntail (and may be WildFly) CDI extension code, as well as any Quarkus code integrating smallrye-jwt be able to reuse the same code.

sberyozkin:
Making `NullJsonWebToken` independent will help clean up a bit some of Quarkus code, while `ProviderInjectionSupport` will help with Thorntail CDI clean up and WildFly might use it too.

Fixes #132.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/cdi/ProviderExtensionSupport.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/cdi/ProviderExtensionSupport.java
@@ -1,0 +1,163 @@
+/**
+ *
+ *   Copyright 2017 Red Hat, Inc, and individual contributors.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/cdi/ProviderExtensionSupport.java
```diff
@@ -0,0 +1,164 @@
+/**
+ *
+ *   Copyright 2017 Red Hat, Inc, and individual contributors.
```
kenfinnigan: I'd remove the copyright header as we have a LICENSE in the project root

Issue #143, opened 2019-11-07
Support elliptic curve ES256 public keys


aaronanderson:
I followed the instructions for [AWS ALB OIDC authentication](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html) and configured Quarkus/SmallRye to read in the JWT from an HTTP header in the MP application.properties file:

```
#mp.jwt.verify.publickey.location=https://public-keys.auth.elb.us-west-1.amazonaws.com/some-uuid
mp.jwt.verify.publickey.location=META-INF/aws-jwt.pem
mp.jwt.verify.issuer=https://someorg.okta.com
smallrye.jwt.token.header=X-Amzn-Oidc-Data
quarkus.smallrye-jwt.auth-mechanism=MP-JWT
quarkus.smallrye-jwt.enabled=true

```
For whatever reason AWS resigns the JWT with their own key and they use Elilptic Curve based encryption and not RSA. When SmallRye tries to parse the PEM file remotely or locally it only tries to decode the PEM file as an RSA public key and fails.

starksm64:
For reference, info on the signature and header from:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html#user-claims-encoding

The JWT signature is ECDSA + P-256 + SHA256.
The JWT header is a JSON object with the following fields:
```json
{
   "alg": "algorithm",
   "kid": "12345678-1234-1234-1234-123456789012",
   "signer": "arn:aws:elasticloadbalancing:region-code:account-id:loadbalancer/app/load-balancer-name/load-balancer-id", 
   "iss": "url",
   "client": "client-id",
   "exp": "expiration"
}
```

Issue #126, opened 2019-08-23
Support the symmetric signature verification


sberyozkin:
It can be possible that the the server itself creates and signs the original JWT token to be used later for the cookie-based authentication (ex, JHipster). To access such tokens via MP JWT Api smallrye-jwt needs to optionally support the hash HS* algorithms which are currently blocked and then a user would just set a whitelist property enabling HS*.
The major change would have to be done in `KeyLocationResolver`. It will only be supported for the keys stored in the JWK format, and the `verificationKey` property would have to become just a `Key` type. 
The somewhat unfortunate bit is that the location property is called `mp.jwt.publickey.location`, so a `smallrye.jwt.secretkey.location` would likely need to be introduced (it is unlikely the secret keys will be supported at the spec level)

gustf:
Hi, do you have any information if and when this will be prioritized?

sberyozkin:
@gustf Hi, I'll try to look into it asap

gustf:
@sberyozkin awesome! 👍

gustf:
@sberyozkin any news regarding this issue?

sberyozkin:
@gustf Sorry, missed your ping. The initial support will be available in the form in `smallrye-jwt-2.2.0` directly from the injectable `JWTParser`, but more work will need to do a solution which will work for both verification and decryption symmetric keys.

gustf:
ok thanks!

PR #145, opened 2019-11-07
fixed Elliptic Curve PEM decoding problem

aaronanderson:
Update to support ES256 PEM files reported in issue [143](https://github.com/smallrye/smallrye-jwt/issues/143)

I needed to add ES256 to the smallrye.jwt.whitelist.algorithms property in the application.properties file in order to successfully verify ES256 signed JWT tokens.


sberyozkin:
@aaronanderson thanks for this PR. @starksm64 initiated a discussion about it on the Smallrye forum. Let me propose for this PR be enhanced further based on that discussion so far.

First of all, we did add a `whitelist` configuration property as part of supporting the customization of how smallrye-jwt parses the tokens, but as your PR shows it was actually of no help to anyone who'd want to try ES256 for ex. Also my understanding is that it is very unlikely that several algorithm types or keys of the different size will be used for a single endpoint, which makes this property even more redundant, for now at least.

So here is a proposal:
* Introduce `io.smallrye.jwt.SignatureAlgorithm` which for now will have 6 values, RS256/384/512 and the same for ES*
* Introduce `io.smallrye.jwt.verify.algorithm` of `SignatureAlgorithm` type and use its value to set a whitelist property, default should be `SignatureAlgorithm.RS256'.. 
* Deprecate `smallrye.jwt.whitelist.algorithms` with a note that `io.smallrye.jwt.verify.algorithm` should be used to customize the algorithm. In fact given that it could not be used before your PR for ES* I'd actually delete it.
* Drop the fallback try from `KeyUtils.decodePublicKey(String content)`, add instead `KeyUtils.decodePublicKey(String content, SignatureAlgorithm algo)`.
* Update both `JWTAuthContextInfoProvider` and `KeyLocationResolver` to call `KeyUtils.decodePublicKey(String content, SignatureAlgorithm algo)`.

That should be it, quite a few steps but they are all very simple :-). Give it a try please
CC @starksm64

aaronanderson:
Hi @sberyozkin I updated the pull request as requested. I also added HMAC JWT signature validation support.

sberyozkin:
@aaronanderson thanks for finding the time to continue working on this PR. I would've preferred if we had dealt with the HMAC as part of #126 - the symmetric signature verification is somewhat off topic for smallrye-jwt. I'll comment on the related parts in your PR, but ideally you'd follow up with another PR for #126. thanks

implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
```diff
@@ -46,42 +51,58 @@
  */
 public class KeyUtils {
     private static final String RSA = "RSA";
+    private static final String EC = "EC";
 
-    public static PrivateKey readPrivateKey(String pemResName) throws IOException, GeneralSecurityException {
+    public static PrivateKey readPrivateKey(String pemResName, SignatureAlgorithm algo)
```
sberyozkin: @aaronanderson please retain all the original signatures and delegate to the overloaded ones with the RS256. Otherwise some code, starting from the test code in Thorntail, will fail to compile

implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
```diff
@@ -118,31 +138,51 @@ public static PublicKey decodeJWKSPublicKey(String jwksValue) throws GeneralSecu
             // A JWK
             jwk = jwks;
         }
-        String e = jwk.getString("e");
```
sberyozkin: @aaronanderson Thanks for taking on this legacy code as well. Instead of manually creating RS and EC keys here, please copy the code from `KeyLocationResolver` which reads JSON into JWK and returns a PublicKey

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -53,23 +53,20 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
             JwtConsumerBuilder builder = new JwtConsumerBuilder()
                     .setRequireExpirationTime();
 
-            if (authContextInfo.getWhitelistAlgorithms().isEmpty()) {
-                builder.setJwsAlgorithmConstraints(
-                        new AlgorithmConstraints(AlgorithmConstraints.ConstraintType.WHITELIST,
-                                AlgorithmIdentifiers.RSA_USING_SHA256));
-            } else {
+            if (!authContextInfo.getWhitelistAlgorithms().isEmpty()) {
```
sberyozkin: @aaronanderson white list should not be empty at this stage. Yes we have deprecated the public property, but internally it should be set at this point to the current signature algorithm

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -53,23 +53,20 @@ public JwtContext parse(final String token, final JWTAuthContextInfo authContext
...
             }
-
             if (authContextInfo.isRequireIssuer()) {
                 builder.setExpectedIssuer(true, authContextInfo.getIssuedBy());
             } else {
                 builder.setExpectedIssuer(false, null);
             }
             if (authContextInfo.getSignerKey() != null) {
                 builder.setVerificationKey(authContextInfo.getSignerKey());
+            } else if (authContextInfo.getHmacSecret() != null) {
```
sberyozkin: This is something I'm not sure about. I doubt we should let the users configure the signing private keys directly in the application.properties. My only plan was to support a secret key location possibly restricting it to HTTPS based JWKs only for a start and may be relaxing it to support the file system.

implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTAuthContextInfo.java
```diff
@@ -213,4 +226,20 @@ public String getGroupsSeparator() {
...
+
+    public void setSignatureAlgorithm(SignatureAlgorithm signatureAlgorithm) {
+        this.signatureAlgorithm = signatureAlgorithm;
+    }
+
+    public SignatureAlgorithm getSignatureAlgorithm() {
+        return signatureAlgorithm;
+    }
+
+    public void setHmacSecret(byte[] hmacSecret) {
```
sberyozkin: Lets isolate the secret key configuration to the location only

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -64,10 +78,11 @@ public static JWTAuthContextInfoProvider createWithKey(String publicKey, String
      * @return a new instance of JWTAuthContextInfoProvider
      */
     public static JWTAuthContextInfoProvider createWithKeyLocation(String publicKeyLocation, String issuer) {
-        return create(NONE, publicKeyLocation, issuer);
+        return create(NONE, publicKeyLocation, NONE, issuer);
     }
 
-    private static JWTAuthContextInfoProvider create(String publicKey, String publicKeyLocation, String issuer) {
+    private static JWTAuthContextInfoProvider create(String publicKey, String publicKeyLocation, String hmacSecret,
```
sberyozkin: IMHO we don't need this overload, as we can't have an option where both the asymmetric and secret keys are used

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -226,6 +246,20 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
+     */
+    @Inject
+    @ConfigProperty(name = "io.smallrye.jwt.verify.algorithm", defaultValue = "RS256")
+    String signatureAlgorithm;
+
+    /**
+     * The HMAC secret if HMAC is used.
+     */
+    @Inject
+    @ConfigProperty(name = "io.smallrye.jwt.verify.hmac-secret")
```
sberyozkin: As suggested at #126, IMHO it should be `secret-key-location`, I'm not also that 'hmac' should be in the name, in case we decide to support some other secret key signature algorithm.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -53,7 +55,19 @@
...
+
+    /**
+     * Create JWTAuthContextInfoProvider with the public key and issuer
+     * 
+     * @param secret - Base64 HMAC shared secret
+     *
+     * @param issuer the issuer
+     * @return a new instance of JWTAuthContextInfoProvider
+     */
+    public static JWTAuthContextInfoProvider createWithHmac(String secret, String issuer) {
```
sberyozkin: IMHO we only need one extra overload, `createWtithSecretKey` which would take `secretKeyLocation` and `issuer` parameters. We might add a support for the inline private keys later...

testsuite/basic/src/test/java/io/smallrye/jwt/TestJsonWebToken.java
```diff
@@ -16,15 +16,29 @@
...
 import java.util.HashMap;
 import java.util.HashSet;
 
 import org.eclipse.microprofile.jwt.Claims;
 import org.eclipse.microprofile.jwt.JsonWebToken;
 import org.eclipse.microprofile.jwt.tck.util.TokenUtils;
+import org.jose4j.jwt.JwtClaims;
 import org.testng.annotations.Test;
 
+import com.nimbusds.jose.JWSAlgorithm;
```
sberyozkin: Please use `Jose4j` only, we are trying to avoid referencing Nimbus

sberyozkin:
@aaronanderson I've finished another round of review, in general it is definitely a step toward the better support of various algorithms, thanks, but I'd like to propose that you minimize this PR only about the EC key support as all these changes are quite sensitive, and then follow with another small PR for the secret keys, we can continue with the current PR for all the signature types but oit will just take us longer to resolve it :-)
Thanks

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: The other thing a `.verify.` fragment might be redundant, because for the JWT signing utilities it would also be a signing secret key loation

sberyozkin:
@aaronanderson I'm reworking the PR due to a number of conflicts and to make things a bit simpler, I'd like to take care of the symmetric keys in the next round, too many ambiguities with them so I'd like to take it step by step....

PR #147, opened 2019-11-21
Cache JWTCallerPrincipalFactory within @ApplicationScoped bean.

darranl:
Integrating SmallRye components into the WildFly application server the integrated components are added to the server as a module, within WildFly each module is loaded within it's own ClassLoader which is subsequently shared across deployments.

SmallRye JWT has been making use of a static variable to cache the JWTCallerPrincipalFactory, this has a side effect of caching application specific configuration and CDI beans.  For single deployment processes or processes restarted for each redeployment the static emulates @ApplicationScoped but in WildFly it is a global scope.

This pull request is to propose moving the caching into an @ApplicationScoped bean so in all existing scenarios the behaviour should remain the same.

The existing behaviour of JWTCallerPrincipalFactory.setInstance(JWTCallerPrincipalFactory resolver); has been retained but if desirable an additional method could be added that takes Supplier<JWTCallerPrincipalFactory> which could return a new instance per application.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
@@ -19,7 +19,7 @@ import org.eclipse.microprofile.jwt.JsonWebToken;
 import org.jboss.logging.Logger;
 
 import io.smallrye.jwt.auth.principal.JWTAuthContextInfo;
-import io.smallrye.jwt.auth.principal.JWTCallerPrincipalFactory;
+import io.smallrye.jwt.auth.principal.JWTParser;
 import io.smallrye.jwt.auth.principal.ParseException;
 
 /**
@@ -39,9 +39,11 @@ public abstract class AbstractBearerTokenExtractor {
     private static final Logger LOGGER = Logger.getLogger(AbstractBearerTokenExtractor.class);
 
     private final JWTAuthContextInfo authContextInfo;
+    private final JWTParser jwtParser;
 
-    protected AbstractBearerTokenExtractor(JWTAuthContextInfo authContextInfo) {
+    protected AbstractBearerTokenExtractor(JWTAuthContextInfo authContextInfo, JWTParser jwtParser) {
         this.authContextInfo = authContextInfo;
+        this.jwtParser = jwtParser;
     }
 
     /**
@@ -121,8 +123,7 @@ public abstract class AbstractBearerTokenExtractor {
     }
 
     public JsonWebToken validate(String bearerToken) throws ParseException {
-        JWTCallerPrincipalFactory factory = JWTCallerPrincipalFactory.instance();
-        return factory.parse(bearerToken, authContextInfo);
+        return jwtParser.parse(bearerToken);
     }
 
     /**
--- a/implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
@@ -19,8 +19,6 @@ import org.eclipse.microprofile.jwt.JsonWebToken;
 import org.jboss.logging.Logger;
 
 import io.smallrye.jwt.auth.principal.JWTAuthContextInfo;
-import io.smallrye.jwt.auth.principal.JWTParser;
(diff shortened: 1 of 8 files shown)


kenfinnigan:
Possibly a typo:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc (default-cli) on project smallrye-jwt: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /home/runner/smallrye-jwt/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java:25: error: reference not found
[ERROR]  * A default implementation of {@link JWRParser}.
```

implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
```diff
@@ -39,9 +39,11 @@
     private static final Logger LOGGER = Logger.getLogger(AbstractBearerTokenExtractor.class);
 
     private final JWTAuthContextInfo authContextInfo;
+    private final JWTParser jwtParser;
 
-    protected AbstractBearerTokenExtractor(JWTAuthContextInfo authContextInfo) {
+    protected AbstractBearerTokenExtractor(JWTAuthContextInfo authContextInfo, JWTParser jwtParser) {
         this.authContextInfo = authContextInfo;
+        this.jwtParser = jwtParser;
```
sberyozkin: @darranl Can this parser be injected ? Otherwise we'd need to change Thorntail and Quarkus code a bit to pass it around as well, so the `smallrye-jwt` version will likely have to changed to 3.0.0 - not a problem in itself, but please check

darranl: @sberyozkin I will have a look, at the moment I just injected where we already have injection occurring, I think the final place it is used wasn't a CDI bean.

sberyozkin: @darranl I've looked at Quarkus/Thorntail code, not sure we can avoid passing `JWTParser` to the token extractors, but I'm also not sure either now `JWTParser` needs to be passed to them :-), it is only used indirectly [here](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/jaxrs/JWTAuthenticationFilter.java#L64), which is where it is actually have to be used directly because the extractor is not really about the validation. So please make that `validate` method in the abstract extractor code deprected and use the parser directly in the JAX-RS filter which is where is it already injected to in your PR and it should do it

Thanks

sberyozkin: @darranl In fact I'd just remove that `valdate` method since the only consumer of that method is this JAX-RS filter AFAIK. But I'm OK if we stay on the safer side and only deprecate it. CC @MikeEdgar Hi Mike, FYI

MikeEdgar: @sberyozkin - +1 to removing `validate`. I believe it was originally copied from Thorntail and made public so that it was callable from the JAX-RS filter and the HTTP auth mechanism class.

sberyozkin:
@darranl Darran, can you please do a minor update as suggested above ?

darranl:
@sberyozkin just going to look at the Optional claim injection issue we were discussing so I can get a PR in for that one and then will come and look for a solution for this one.

At that point that should be the three known issues needed for the WildFly integration, it looks like we are almost 100% on the TCK now.

darranl:
Going to look at changing this one now.

darranl:
Ok I have just pushed another commit to this branch, is this what you intended?  I haven't performed a WildFly / TCK run yet but will do this afternoon along with the next changes for Optional claims.

darranl:
With the other changes submitted so far and the latest in this PR within WildFly the JWT TCK is now passing 100%

sberyozkin:
@darranl good news. Sorry for a delay, I need to get to some conclusion on the builder APi work asap. Will test your PR immediately afterwards. Thanks

darranl:
Thank you @sberyozkin

darranl:
@sberyozkin if this one gets merged would it also be possible to request a tag of SmallRye JWT - once we have a tag with these changes we are ready to start some of the merges into WildFly.

PR #155, opened 2019-12-11
Support typed Optional claim injection as required by the specification and TCK.

darranl:


implementation/src/main/java/io/smallrye/jwt/auth/cdi/RawClaimTypeProducer.java
```diff
@@ -121,18 +121,14 @@ Double getClaimAsDouble(InjectionPoint ip) {
...
+     * Produces a *raw* Optional value.
      *
      * @param ip reference to the injection point
      * @return an optional claim value
      */
     @Produces
     @Claim("")
-    @SuppressWarnings("rawtypes")
-    public Optional getOptionalValue(InjectionPoint ip) {
+    public <T> Optional<T> getOptionalValue(InjectionPoint ip) {
```
sberyozkin: Hi Darran @darranl as far as I recall that fails in Thorntail, I've tried it before, can conform tomorrow. Perhaps we can reuse the Quarkus idea somehow instead if it proves a problem...

darranl: The Quarkus approach seems to rely on an io.quarkus.arc.BeanCreator so that seems Quarkus specific, unless there is a pure CDI equivalent we can use?

I do have a project I could add my own producer to but that is more of a last resort, ideally I think SmallRye JWT should contain the producer and the current definition is not compatible with the CDI specification - however maybe there is an alternative.

For the various feedback I have received on the PRs is feels like maybe some central customisation within the CDI extension could help to provide "tweaks" to the capabilities of SmallRye JWT.

We could either make use of an abstract Extension or add some additional methods to the existing extension but then a child extension could override specific behaviour.  As an example we could have a method: -

protected boolean useUntypedOptionalClaims() {
    return true;
}

Where this returns true the extension can register a producer of the original format and if false register a producer with this new format.

This is a bit of a tangent for this PR but there are other methods that could be useful to override such as isHttpAuthMechanismEnabled() and isEESecurityAvailable() - in terms of the WildFly integration I would hard code both of these to true but by default we would leave the existing behaviour.

Going back to my earlier PR regarding the static vs @ApplicationScoped bean, extension overrides may be a way to better handle that.

Sorry that did diverge quite a bit ;-)

sberyozkin: @darranl What is interesting it is actually working in Thorntail now :-), @Ladicek updated it to WildFly 18 and I guess it has made all the difference. Checking Quarkus now...

Re your PR #147, yes, lets discuss the other options there (or at the related issue). I was somewhat concerned (without exactly understanding why :-) ) about introducing a new `entity` into the API, such as `JwtParser` - may be because now there would be two ways for the consumers to bootstrap `smallrye-jwt`, via the factory or via this new injected parser. (FYI, in Quarkus, at some point there was some duplication, no factories were used and the code was paste/copied so I move it all into `DefaultJwtParser`). #147 still looks good to me though...

darranl: +1 on continuing on #147 - if we can find a solution for this one I think #147 is my final issue to have a clean TCK run with SmallRye JWT integrated directly into WildFly.

sberyozkin: Well, Quarkus does not like it. Let me ask @mkouba

darranl: With Quarkus I wonder if it conflicts with their workaround?

mkouba: I'm sorry what's the question?

sberyozkin: Hi Martin @mkouba , was just about to ping you on Zulip dev :-), so with Darran's fix I'm seeing:
```
Suppressed: javax.enterprise.inject.AmbiguousResolutionException: Ambiguous dependencies for type java.util.Optional<javax.json.JsonString> and qualifiers [@Claim(standard = preferred_username)]
	- java member: io.quarkus.jwt.test.DefaultScopedEndpoint#currentUsername
	- declared on CLASS bean [types=[io.quarkus.jwt.test.DefaultScopedEndpoint], qualifiers=[@Default, @Any], target=io.quarkus.jwt.test.DefaultScopedEndpoint]
	- available beans:
		- PRODUCER METHOD bean [types=[java.util.Optional<javax.json.JsonString>, java.lang.Object], qualifiers=[@Any, @Claim(value = "")], target=java.util.Optional<javax.json.JsonString> getOptionalJsonString(javax.enterprise.inject.spi.InjectionPoint ip), declaringBean=io.smallrye.jwt.auth.cdi.JsonValueProducer]
		- PRODUCER METHOD bean [types=[java.util.Optional<T>, java.lang.Object], qualifiers=[@Any, @Claim(value = "")], target=java.util.Optional<T> getOptionalValue(javax.enterprise.inject.spi.InjectionPoint ip), declaringBean=io.smallrye.jwt.auth.cdi.RawClaimTypeProducer]
		at io.quarkus.arc.processor.Beans.resolveInjectionPoint(Beans.java:487)
		at io.quarkus.arc.processor.BeanInfo.init(BeanInfo.java:416)
		at io.quarkus.arc.processor.BeanDeployment.init(BeanDeployment.java:215)
```
which is probably correct if Quarkus also registers an explicit producer for JSON..., may be somewhere here:

https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-jwt/deployment/src/main/java/io/quarkus/smallrye/jwt/deployment/SmallRyeJwtProcessor.java#L123

FYI, @darranl would like to have something like this supported:
```
// not only JSON String:
@Inject
@Claim(standard = Claims.birthdate)
Optional<JsonString> birthdate;

// or, just:
@Inject
@Claim(standard = Claims.birthdate)
Optional<String> birthdate;

```

sberyozkin: @mkouba I wonder if we can move some CDI code out of Quarkus to smallrye-jwt to avoid the conflicts. FYI, Darran's fix is OK in the latest Thorntail

sberyozkin: @mkouba actually Thorntail does not have a single test checking `JsonString`, let me check that too

mkouba: Yes, it's correct because according to the typesafe resolution rules a bean with type `Optional<T>` is assignable to the `Optional<JsonString>` required type:
```
A parameterized bean type is considered assignable to a parameterized required type if they have identical raw type and for each parameter:
* the required type parameter is an actual type, the bean type parameter is a type variable and the actual type is assignable to the upper bound, if any, of the type variable,
```
In this case, `JsonString` is assignable to `java.lang.Object`.

For a raw type (`Optional`) a different rule applies:
```
A raw bean type is considered assignable to a parameterized required type if the raw types are identical and all type parameters of the required type are either unbounded type variables or java.lang.Object.
```

mkouba: @manovotn Pls do a quick sanity check ^ ;-)

mkouba: @sberyozkin it should NOT work in TT!

sberyozkin: @mkouba Hi Martin, so this is why I'm thinking aloud that we should have a piece of CDI code in `smallrye-jwt` which can be reused in Quarkus/WildFly/Thorntail, in Quarkus this would be shared code would be used one way, in WildFly - as part of some CDI extension, etc.

mkouba: The problem is that we can't reuse the CDI extensions in Quarkus. Of course, producers and similar stuff could be reused.

darranl: Just to clarify, this is not that I would like Optional<String> claim injection supported,  I was going to try and adapt the quickstart to avoid that if it is not presently possible,

However the MicroProfile JWT TCK specifically tests typed Optional claim injection and as it is defined in the specification I don't think we can challenge the TCK.

manovotn: > In this case, JsonString is assignable to java.lang.Object.

Yes, that's exactly the case.

> For a raw type (Optional) a different rule applies:

Yeah, basically `Optional` can be assigned to something like `Optional<T>` and `Optional<Object>`

sberyozkin: @mkouba yeah, we'd only have something similar to [ProviderExtensionSupport](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/cdi/ProviderExtensionSupport.java). In Thorntail the CDI extension extends it, but in Quarkus it can be used as a helper utility code. So we'd have some shared code which would be reused somehow differently in different projects.

sberyozkin: @mkouba @manovotn Hi Martin, Matej sure, this is now clear :-). So is there a chance some code which makes it work in Quarkus can be moved to `smallrye-jwt` ? @darranl then can update the PR accordingly.
Martin, yes, it does not work in Thorntail with `<T>`, you are right :-)

darranl: Which code are we saying can be moved from Quarkus?  The code I have seen is making use of Quarkus APIs to add this support.

sberyozkin: @darranl I'm not sure, but yes, there is Quarkus API there but it seems some of that code is not tied to Quarkus and can be moved to a `smallrye-jwt` utility class, for it be reused from Quarkus APi or CDI extensions in WildFly/Thorntail. Can you something that might be moved from  [here](https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-jwt/deployment/src/main/java/io/quarkus/smallrye/jwt/deployment/SmallRyeJwtProcessor.java#L123) and/or [here](https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-jwt/runtime/src/main/java/io/quarkus/smallrye/jwt/runtime/auth/RawOptionalClaimCreator.java#L17) ?

darranl: It sounds to me the main aspect is ensuring Optional, Optional<?>, Optional<String + other types>, Optional<JsonString>.

If Quarkus is reporting a "Ambiguous dependencies" error it sounds like the duplicate registration would then need to be removed from within Quarkus.

darranl: Looks like I can reproduce without Quarkus if I add all the Optional types to my Quickstart.

darranl: Outside of Optional<JsonValue> there are only four types required to be supported by the spec String, Set<String>, Long, and Boolean - rather than a generic producer I suspect easier to just add each individual one.

darranl:
Lets try an alternative commit.

I was able to reproduce the ambiguity injecting Optional<JsonString> within WildFly as well, as there are only four claim types required by the specification I have added individual @Produces methods for each type.

For now I have also left in the method for a raw Optional - this should not really be there as the spec does not require it however I believe due to it's prior existence deployments could have made use of it so maybe safer to leave it in place for now.

Locally with these changes I am still at 100% running the TCK within WildFly and also able to deploy my quickstart with various permutations of Optional<?> being correctly resolved.

darranl:
Any clue as to what the failure is?  I suspect it doesn't like something in the Javadoc but I can't see an actual error.

sberyozkin:
@darranl try `mvn verify javadoc:javadoc`

darranl:
Now I definitely don't know that the failure is, I did have one issue with mvn verify javadoc:javadoc that I have fixed but now I get a clean run locally.

sberyozkin:
@darranl this now fails in Thorntail with a very similar issue to Quarkus: 

```
2019-12-13 12:28:35,304 ERROR [io.undertow.request] (default task-1) UT005023: Exception handling request to /mpjwt/subject/secured/provider-optional: org.jboss.resteasy.spi.UnhandledException: org.jboss.weld.exceptions.AmbiguousResolutionException: WELD-001335: Ambiguous dependencies for type Optional<String> with qualifiers @Claim
 Possible dependencies: 
  - Producer Method [Optional] with qualifiers [@Claim] declared as [[UnbackedAnnotatedMethod] @Produces @Claim public io.smallrye.jwt.auth.cdi.RawClaimTypeProducer.getOptionalValue(InjectionPoint)],
  - Producer Method [Optional<String>] with qualifiers [@Any @Claim] declared as [[UnbackedAnnotatedMethod] @Produces @Claim public io.smallrye.jwt.auth.cdi.RawClaimTypeProducer.getOptionalString(InjectionPoint)]
```

while handling [this point](https://github.com/thorntail/thorntail/blob/master/testsuite/testsuite-microprofile-jwt/src/main/java/org/wildfly/swarm/microprofile/jwtauth/ApplicationScopedSubjectExposingResource.java#L54).

FYI, Thorntail reuses this [ProviderExtensionSupport](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/cdi/ProviderExtensionSupport.java).

@mkouba @manovotn hey Martin, Matej, I do feel now it would be good to remove the non-Arc specific code from Quarkus smallrye-jwt which deals with CDI into some utility code in this project to make it easy for this PR to go ahead. no CDI code is left in Thorntail now (apart from the nearly empty CDI extension extending [ProviderExtensionSupport](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/cdi/ProviderExtensionSupport.java) and registering the producers).

sberyozkin:
@darranl though may be just adding some extra fix to [ProviderExtensionSupport](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/cdi/ProviderExtensionSupport.java) will do...

darranl:
TBH I am not quite sure what else is needed, in the new year I think it may be an idea to look at defining some scope around SmallRye JWT as on pull requests submitted so far the bigger discussion points seem to be around the integrations with other servers.  A couple of points I think I would suggest looking into are: -

1. Definitions of public / private API / SPI - as a new contributor it is not clear what is public to allow cross package access and what is public for external use.
2. Defining the CDI version, and triaging contributions in relation to the specification.

For now it is not really practical for me to stop what I am doing to start debugging Thorntail so I don't really understand what change would be required to ProviderExtensionSupport - for now I am primarily making use of SmallRyeJWTAuthCDIExtension which doesn't in turn make use of ProviderExtensionSupport so that one seems specific to Thorntail at the moment?

At this point it feels like both Quarkus and WildFly will run with the new producer methods, I think it will be better if I go back and look at the actual CDI Extension and I will add some way to control which annotated type is registered.  The present type can be left as it is so should retain backwards compatibility provided they are not all dynamically discovered, the present type can also be marked as deprecated.  A new type will be added with the new Optional producer methods, WildFly can use that now, Quarkus sounds like only a small step to migrate to it once the ARC specifics are removed and Thorntail can move when ready.

Finally once each integration has migrated the deprecated code will be deleted from the project.

Overall I believe SmallRye JWT is presently in use with five runtimes if we include WildFly?  Sounds like it would make sense to introduce code in such a way that the individual runtimes can choose when to activate it rather than attempt to test and fix them all at once.

sberyozkin:
@darranl Do you do anything special in WildFly to support `Provider` typed injections ? This `ProviderExtensionSupport` is the code which was internal to the Thorntail CDI extension at some point but I've extracted it into `ProviderExtensionSupport` and moved to `smallrye-jwt` thinking you may need it for the WildFly integration work :-)

darranl:
@sberyozkin At this point once I add those new Optional<?> methods I have complete support for the types required by the TCK / Spec and TBH I think they look easier to read.

I can have a look at if the addTypeToClaimProducer could work but I would probably have to extend SmallRyeJWTAuthCDIExtension to see about activating it.  I can't see this code in action at the moment but if I am reading it correctly it supports Optional<AnyType> wheras the spec defined a specific list of types that should be supported.

sberyozkin:
@darranl I have just confirmed that Quarkus also fails but we are discussing the options with Martin

darranl:
Looking into the MPJWTExtension in Thorntail each producer is registered individually: -

```
for (Class<?> clazz : new Class<?>[] {
                JWTAuthContextInfoProvider.class,
                CommonJwtProducer.class,
                PrincipalProducer.class,
                RawClaimTypeProducer.class,
                ClaimValueProducer.class,
                JsonValueProducer.class,
        }) {
            bbd.addAnnotatedType(beanManager.createAnnotatedType(clazz), extensionName + "_" + clazz.getName());
        }
```

If I was to add my new methods to a new class OptionalRawClaimTypeProducer then Thorntail could continue to use the existing class which we know works, Quarkus would continue to use the existing class which we know works and WildFly could register the new class.

If you are saying Quarkus fails I think the approach that allows controlled / incremental adoption is safer all round.

sberyozkin:
@darranl Sounds good, though please let me check first if I can remove the provider extension support and the CDI code in Quarkus with your PR...

darranl:
Ok @sberyozkin I will go back to my oldest PR in the queue as that does still need some work, I am fairly confident now we can come up with some strategy on this one.

sberyozkin:
@darranl Just removing the code in Quarkus/Thorntail did not work for me, so indeed, lets go ahead with your idea of introducing `OptionalRawClaimTypeProducer` which will let us resolve this PR and then later I may ask Martin to help to somehow consolidate the existing `RawClaimTypeProducer` and this new `OptionalRawClaimTypeProducer`

darranl:
The commit I have now added to this PR splits it out into a separate Producer to allow control as to if these methods are registered.

As I have just updated both my PRs I will now rebuild WildFly and run the TCK again.

PR #160, opened 2020-01-10
It should be possible to specify a JWK set of more than one key using the `mp.jwt.verify.publickey` config property.

darranl:
Reviewing the specification there are two different config properties to specify the public key, this can either be contained within the MicroProfile Config or reference an external file.  

The specification defines a complete set of key representations that can be supported which I believe applies equally to both properties so this pull request makes it possible to handle multiple keys from a JWKS specified within the config.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
+++ b/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
@@ -22,7 +22,6 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.UncheckedIOException;
-import java.math.BigInteger;
 import java.security.GeneralSecurityException;
 import java.security.KeyFactory;
 import java.security.KeyPair;
@@ -32,14 +31,20 @@ import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.cert.CertificateFactory;
 import java.security.spec.PKCS8EncodedKeySpec;
-import java.security.spec.RSAPublicKeySpec;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collections;
+import java.util.List;
 
 import javax.json.Json;
 import javax.json.JsonArray;
 import javax.json.JsonObject;
 import javax.json.JsonReader;
+
+import org.jose4j.json.JsonUtil;
+import org.jose4j.jwk.JsonWebKey;
+import org.jose4j.lang.JoseException;
 
 /**
  * Utility methods for dealing with decoding public and private keys resources
@@ -69,7 +74,7 @@ public class KeyUtils {
 
     /**
      * Decode a PEM RSA private key
-     * 
+     *
      * @param pemEncoded - pem string for key
      * @return RSA private key instance
      * @throws GeneralSecurityException - on failure to decode and create key
@@ -86,14 +91,13 @@ public class KeyUtils {
     }
 
     /**
-     * Decode a JWK(S) encoded public key string to an RSA PublicKey. This assumes a single JWK in the set as
-     * only the first JWK is used.
-     * 
-     * @param jwksValue - JWKS string value
-     * @return PublicKey from RSAPublicKeySpec
-     * @throws GeneralSecurityException when RSA security is not supported or public key cannot be decoded
+     * Decode a JWK(S) encoded public key string to a list of JsonWebKeys.
+     *
+     * @param jwksValue
+     * @return a list of decoded JsonWebKey instances.
+     * @throws GeneralSecurityException when the JsonWebKey can not be decoded.
      */
-    public static PublicKey decodeJWKSPublicKey(String jwksValue) throws GeneralSecurityException {
+    public static List<JsonWebKey> decodeJsonWebKeySet(String jwksValue) throws GeneralSecurityException {
         JsonObject jwks;
 
         try (Reader reader = new StringReader(jwksValue);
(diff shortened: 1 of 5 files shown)


sberyozkin:
@darranl what text you are referring to ? I'm not sure yet about all these changes, they will conflict with the builder API PR, and I'd also like to avoid leaking Jose4j API into the context info.
The other thing I'm not sure about introducing many resolvers, it took me awhile to bring the resolution code spread across the code base into a single class.

darranl:
@sberyozkin The text in section 9.2 of the MicroProfile JWT 1.1 specification, it specifies that public keys will be represented in one of the following formats: -

• Public Key Cryptography Standards #8 (PKCS#8) PEM
• JSON Web Key (JWK)
• JSON Web Key Set (JWKS)
• JSON Web Key (JWK) Base64 URL encoded
• JSON Web Key Set (JWKS) Base64 URL encoded 

Within the specification I don't find any further text specifying that either of the config values may choose to only support a subset of this list.

Within the JWKS documentation I also do not find any text that describes just dropping back to support only the first key in the set.

sberyozkin:
@darranl thanks, given that IMHO we really should not try to overload what 'publickey' alone can refer to, why would we want to do it ?

darranl:
@sberyozkin TBH I don't think this is the correct place for that discussion.

We have a published specification and an implementation claiming to implement that specification.

As I have said elsewhere I do believe it is time for a round of clarification within the specification and this is definitely the kind of discussion we should include in those discussions, for compatibility purposes probably making some of these optional.

But in the meantime unless the specification is considered more along the lines of a set of guidelines  we need an implementation that implements the detail described within the specification.

By deviating from the specification in this way I have a QE team that don't believe SmallRye JWT is ready for inclusion in WildFly.

sberyozkin:
@darranl Right. Lets sort it out. Let me get the builde API in first though, and then I'll work with you on getting it in before the next smallrye-kwt release. Thanks

darranl:
Thank you @sberyozkin 

Regarding the actual implementation, let me know if you want me to I can look into an alternative implementation if you are worried about the exposed API - another possibility I could see is instead of the list of keys we could pass in a VerificationKeyResolver keeping the details of it's implementation internal.

sberyozkin:
@darranl sure, the builder API has just got merged by Scott, have a look please, but also give me a bit of time today to catch up with what was happening re 'publickey' before, as I've just got back from PTO. I see the public key was retrieved from a method whose name was a bit confusing :-)
Yes, for example, with Thorntail `keycloak-microprofile-jwt` fraction, jose4j would not be be needed,

I wonder if we really should deprecate RSAPublicKey getter from the context info class, and have all the logic inside `KeyLocationResolver` (split internally if needed). This is what `KeyLocationResolver` does right now with the `publickey.location`:
- try JWK set from HTTPS and if that does not work, then get the content from InputStream (which can be the filesystem. classpath, or plain http or non JWK set HTTPS), and once the content is loaded from the stream then there is a master function there which tries all those key types which this content may have

So, for this PR, may be what can be done is this: if 'publickey' is set then we already have a content which can be just passed to the existing code which loads the key from String, whatever it may be.

There is some logic there to support JWK set with a single key only if the token 'kid' is null, as well as an option to cache a JWK key from the set if the user has a set a kid property.

What do you think ?

sberyozkin:
@darranl Hi Darran, I've looked at the code more carefully and I think there may be a short term solution which would avoid any API changes.
In `JWTAuthContextInfoProvider`, if its `decodeMpJwtPublicKey` fails, then we use an `inline:` scheme and do `contextInfo.setPublicKeyLocation("inline:" + mpJwtPublicKey.get())`.

(in fact we can just remove decodeMpJwtPublicKey() and deprecate RSA key getter/setter on the context but it can be reviewed later).
 
And in `KeyLocationResolver.readKeyContent`, if it is an `inline:` scheme then the content is returned as is. So if it will be a JWK set with more than one key then it will be taken care of with the possible optimizations I've mentioned before.

If you agree then lets do it ? I can hack it quickly enough as well, let me know please, as i'd like to get the builder API available to Quarkus users asap

Thanks

darranl:
@sberyozkin that strategy also sounds fine - if we can avoid an API change for now that is probably better.

Not for this PR but one thing may be useful to discuss at a later point is if it could make sense to allow the VerificationKeyResolver to be replaced without needing to replace the complete TokenParser - within the app server at the moment it feels like I could find more benefits to being able to plug in an alternative key resolution to tie into some of our existing resources rather than replacing the parser.

sberyozkin:
@darranl I was thinking this morning that this `inline` workaround is not even needed as there will be some overhead if the `publickey` points to a multi-key JWK set, and what I typed initially should work very well.  We just don't do any key processing in  `JWTAuthContextInfoProvider` (though `decodeMpJwtPublicKey` may need to stay as Thorntail call it I think), and in KeyLocationResolver the `publickey` content is set is preferred.
Do you have time today to look at something like that ?

Re `VerificationKeyResolver`, I think it is fine to introduce it later (perhaps such that it is used if no default `KeyLocationResolver` can find the key, or the other way around, use the defautt one if the custom one can't find it so as to guarantee that the the default resolution policy will be there if needed)

Thanks

darranl:
@sberyozkin I am seeing if I can get to this today but FYI it is our feature freeze today so have a pile of tasks to get through.

I would say don't hold up the SmallRye tag for this if others are waiting for a tag, we may just want a follow up at some point once this can be incorporated.

sberyozkin:
@darranl sounds good, I'll try to merge the other open PR today/tomorrow and do a tag, as the proposed change here is a bit sensitive. I might actually close this PR as the conflict here would not be relevant should the proposed change will be implemented. I might do a new PR too to save you some time, cheers

sberyozkin:
@darranl The more I think about the option of deprecating RSA key getters from the context info the happier I become :-) as I was thinking for a while before if i could do it. It makes sense IMHO to continue consolidating the code to do with the key resolution in one place (with the key resolver customization as per your proposal going forward too). But indeed, lets do it in the follow up tag to avoid any issues...

sberyozkin:
@darranl I have a PR for you to review coming in shortly, so let me close this one

PR #162, opened 2020-01-13
Add support for custom token schemes

acoburn:
Add a `smallrye.jwt.token.schemes` configuration value to support additional token schemes, such as DPoP.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
+++ b/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
@@ -41,6 +41,7 @@ import io.smallrye.jwt.auth.principal.JWTAuthContextInfo;
 @Dependent
 public class JWTAuthContextInfoProvider {
     private static final String AUTHORIZATION_HEADER = "Authorization";
+    private static final String BEARER_SCHEME = "BEARER";
     private static final String NONE = "NONE";
     private static final String DEFAULT_GROUPS_SEPARATOR = " ";
     private static final Logger log = Logger.getLogger(JWTAuthContextInfoProvider.class);
@@ -77,6 +78,7 @@ public class JWTAuthContextInfoProvider {
         provider.tokenHeader = AUTHORIZATION_HEADER;
         provider.tokenCookie = Optional.empty();
         provider.tokenKeyId = Optional.empty();
+        provider.tokenSchemes = Optional.of(BEARER_SCHEME);
         provider.requireNamedPrincipal = Optional.of(Boolean.TRUE);
         provider.defaultSubClaim = Optional.empty();
         provider.subPath = Optional.empty();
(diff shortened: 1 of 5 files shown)


implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -41,6 +41,7 @@
 @Dependent
 public class JWTAuthContextInfoProvider {
     private static final String AUTHORIZATION_HEADER = "Authorization";
+    private static final String BEARER_SCHEME = "BEARER";
```
sberyozkin: @acoburn minor one, can you please rename it to `Bearer`, even though the case insensitive comparison is used.

acoburn: My mistake, it should definitely be `Bearer`

sberyozkin: Np, thanks, I suppose even this all capital case variation can happen in some cases

sberyozkin:
@acoburn It all looks good, but FRY, @darranl 's PR is due to go next so there will be a minor rebase action required once it is in.
By the way, Darran, @darranl, can you please check if you are OK with the stream API code ? I think the one which converts a comma-separated list is OK, it is done only once.
(Aaron, FYI, it is fine but WildFly QE team has detected that some of this API may cause too much GC overhead, and we don't want to make the WildFly QE team unhappy :-) )

darranl:
Actually the streams one is the one I have noticed not from QE :wink: 

Running the WildFly application server in environments with constrained resources it has in the past been found that garbage collection due to per-request object allocation causes a major overhead compromising the potential throughput.

As MicroProfile in general is expected to run within environments with constrained resources it would be good across both implementations of the specs (and within the actual specifications) to ensure any per request object allocation is minimised to the set of Objects that are essential.

The reason Streams are a bit of a target in per request handling is they potentially lead to multiple Objects being instantiated that are left on the heap, often the same code can be accomplished using for loops and primitives without adding Objects requiring GC.

FYI I am saying this as an engineer who liked the new APIs as they became available to us, liked using Streams etc with lambdas and after various performance tests has had to slowly remove these from my own code :smile:.

One time use of Streams is probably not too bad but anything that possibly slows down either a deployment or the processing of a first request should be considered to make sure it is really required.

sberyozkin:
@darranl Thanks :-), sounds good

acoburn:
@darranl thanks for the feedback. I updated the implementation to use regular `for` iterators in place of the Stream API. As a note: there is another use of the Stream API (unrelated to this PR) in the `SmallRyeJwtUtils::setWhitelistAlgorithms` method. Perhaps that should be adjusted in a separate PR.

PR #166, opened 2020-01-16
It should be possible to specify a JWK set of more than one key using the `mp.jwt.verify.publickey` config property

sberyozkin:


sberyozkin:
@darranl Hi Darran, how does this PR look to you ?
thanks

darranl:
Looks good to me, I think I can add approvals in this repo so I will run your changes locally now with my own reproducer and confirm

sberyozkin:
@darranl thanks; this can be further optimized for sure, split across a few verifiers if needed, etc. The problem with RSA key property in the context is that it blocks EC keys so the other open PR already sets it just PublicKey, but then for the jhipster (or indeed any server which creates a JWT cookies with a secret key) that also does not work, so it would have to become `Key`. I hope once we have all concentrated in one place then it would be easier to start cleaning up/optimizing/etc

sberyozkin:
@darranl You've already had many PRs opened so we'll need to deal with the permissions

sberyozkin:
@darranl thank you for initiating this work and spending the time on verifying this PR, and especially for pointing to the fact the `publickey` may hold not only single keys, I was actually quite oblivious of it before I reread the spec :-). I'll merge it a bit later once I test against Quarkus/Thorntail

PR #167, opened 2020-01-20
Add time-to-live config option

### LINKED ISSUE #146: Add a maximum time to live property
The tokens have to be short-lived in general, this is why, in OIDC/OAuth2 flows, refresh tokens with the longer life span exist, to refresh the ATs and thus minimize the theft risk.

JWT has an expiry claim but it may be set for a long time in the future. Sometimes to simplify the client code issuing the tokens. As such there has to be a property which can block the tokens which have been around for too long

MikeEdgar:
Fixes #146 

Also includes migration to Jakarta EE dependencies, supersedes #151 

@sberyozkin - let me know if this aligns with the intent.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
+++ b/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
@@ -85,6 +85,7 @@ public class JWTAuthContextInfoProvider {
         provider.defaultGroupsClaim = Optional.empty();
         provider.groupsPath = Optional.empty();
         provider.expGracePeriodSecs = Optional.of(60);
+        provider.maxTimeToLiveSecs = Optional.empty();
         provider.jwksRefreshInterval = Optional.empty();
         provider.whitelistAlgorithms = Optional.empty();
         provider.expectedAudience = Optional.empty();
@@ -207,6 +208,14 @@ public class JWTAuthContextInfoProvider {
     private Optional<Integer> expGracePeriodSecs;
 
     /**
+     * The maximum number of seconds that a JWT may be issued for use. Effectively, the difference
+     * between the expiration date of the JWT and the issued at date must not exceed this value.
+     */
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.time-to-live.max")
+    Optional<Integer> maxTimeToLiveSecs;
+
+    /**
      * JWK cache refresh interval in minutes. It will be ignored unless the 'mp.jwt.verify.publickey.location' property points
      * to the HTTPS URL based JWK set.
      * Note this property will only be used if no HTTP Cache-Control response header with a positive 'max-age' parameter value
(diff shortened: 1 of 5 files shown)


sberyozkin:
@MikeEdgar Thanks for this PR, I'm not sure it is what I had in mind. We have two options, if it is a 'time to live' kind of restriction then `iat` is required, and I'd indeed call it a `time-to-live` or `maximum-age` property. Or, we can restrict the expiration time, but the `exp` is a date and does not strictly requires `iat`, and as such I reckon if `expiration_limit` is set then we should check that the `exp` date is less then the expiration limit date ? We can support both options but I suppose lets start from the latter one.
And is there a chance you can have have a different PR for updating to Jakarta EE ? (I'm not sure for ex why servlet api is there)
Thanks

MikeEdgar:
@sberyozkin - my thoughts on checking only `exp` in relation to the time of validation is that tokens that are invalid one point in time would become valid at some point in the future (unless I'm misunderstanding).

I'll open a separate PR for the dependencies. I think Servlet is used by `JWTHttpAuthenticationMechanism `.

sberyozkin:
Hi Mike @MikeEdgar this is a really good point, I've been effectively thinking in terms of a single token. So your PR is actually spot on, but we are really talking about the maximum token age, time to live here. Can you please rename it then to `time-to-live` (the internal check in your PR is actually called this way :-) ) ?

MikeEdgar:
> Can you please rename it then to `time-to-live` (the internal check in your PR is actually called this way :-) ) ?

@sberyozkin - you caught me :-) - I renamed it a few times and missed some of the references. It should now refer to TTL/timeToLive throughout the set of changes. The Jakarta EE changes have been moved to #169

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -206,6 +207,14 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
     @ConfigProperty(name = "smallrye.jwt.expiration.grace", defaultValue = "60")
     private Optional<Integer> expGracePeriodSecs;
 
+    /**
+     * The maximum number of seconds that a JWT may be issued for use. Effectively, the difference
+     * between the expiration date of the JWT and the issued at date must not exceed this value.
+     */
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.time-to-live.max")
+    Optional<Integer> maxTimeToLiveSecs;
```
sberyozkin: Lets have it as `Optional<Long>` just in case (see Jose4J `NumericDate`)

sberyozkin:
Hi Mike, @MikeEdgar, looks very good, I've commented asking for one cosmetic change only

sberyozkin:
Sorry @MikeEdgar one more and last suggestion :-), lets just have as `time-to-live`, the extra `.max` bit looks a bit unusual IMHO

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
MikeEdgar: Updates pushed

PR #178, opened 2020-01-31
Support the claim paths with the namespaces

### LINKED ISSUE #164: smallrye.jwt.path.groups and smallrye.jwt.path.sub does not work with namespaced keys containing "/"
Auth0  does not automatically add the sub and groups claims but custom claims can be added.   https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims  Following OIDC specification that advises ensuring custom claims are made globally unique they added a hard requirement that  custom claims to be created with a URI based namespace.  Unfortunately this means that the custom claim for groups and sub if added will always contain a "/" character.  e.g. claim could be "http://mydomain.com/groups" . The code in DefaultJWTTokenParser to find groups based upon the  smallrye.jwt.path.groups  configuration will always split the key value specified as separate path elements and there is currently no way to escape the "/" in these path configuration values with a simple string split.

private List<String> findGroups(JWTAuthContextInfo authContextInfo, JwtClaims claimsSet) {
        if (authContextInfo.getGroupsPath() != null) {
            final String[] pathSegments = authContextInfo.getGroupsPath().split("/");

Options are 
1 - to allow for escaping of / in configuration parameter to split correctly,  
2 - allow for absolute specification of claim key or 
3 - add customization of namespace prefix that will be stripped off claim key before being processed. 
4 - Ability to configure a different separator for the path configuration,  this may get confusing with the smallrye.jwt.groups-separator  configuration used for the group value.  

If a decision is made on the way forward I am happy to have a go implementing this.

sberyozkin:
Fixes #164. 

CC @spbolton

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.eclipse.microprofile.jwt.Claims;
 import org.jboss.logging.Logger;
@@ -42,6 +43,12 @@ import org.jose4j.lang.UnresolvableKeyException;
 public class DefaultJWTTokenParser {
     private static final Logger LOGGER = Logger.getLogger(DefaultJWTTokenParser.class);
     private static final String ROLE_MAPPINGS = "roleMappings";
+    /**
+     * This pattern uses a positive lookahead to split an expression around the forward slashes
+     * ignoring those which are located inside a pair of the double quotes.
+     */
+    private static final Pattern CLAIM_PATH_PATTERN = Pattern.compile("\\/(?=(?:(?:[^\"]*\"){2})*[^\"]*$)");
+
     private volatile VerificationKeyResolver keyResolver;
 
     public JwtContext parse(final String token, final JWTAuthContextInfo authContextInfo) throws ParseException {
@@ -141,7 +148,7 @@ public class DefaultJWTTokenParser {
 
     private String findSubject(JWTAuthContextInfo authContextInfo, JwtClaims claimsSet) {
         if (authContextInfo.getSubjectPath() != null) {
-            final String[] pathSegments = authContextInfo.getSubjectPath().split("/");
+            final String[] pathSegments = splitClaimPath(authContextInfo.getSubjectPath());
             Object claimValue = findClaimValue(authContextInfo.getSubjectPath(), claimsSet.getClaimsMap(), pathSegments, 0);
             if (claimValue instanceof String) {
                 return (String) claimValue;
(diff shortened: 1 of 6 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -42,6 +43,8 @@
 public class DefaultJWTTokenParser {
     private static final Logger LOGGER = Logger.getLogger(DefaultJWTTokenParser.class);
     private static final String ROLE_MAPPINGS = "roleMappings";
+    private static final Pattern CLAIM_PATH_PATTERN = Pattern.compile("\\/(?=(?:(?:[^\"]*\"){2})*[^\"]*$)");
```
MikeEdgar: Just a preference of mine, but it might be helpful for future reading of this regex to have a short comment describing the pattern.

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -141,7 +144,7 @@ private void checkNameClaims(JwtContext jwtContext) throws InvalidJwtException {
 
     private String findSubject(JWTAuthContextInfo authContextInfo, JwtClaims claimsSet) {
         if (authContextInfo.getSubjectPath() != null) {
-            final String[] pathSegments = authContextInfo.getSubjectPath().split("/");
+            final String[] pathSegments = CLAIM_PATH_PATTERN.split(authContextInfo.getSubjectPath());
```
MikeEdgar: Would it be better to do the `split` when the configuration is loaded rather than per request?

sberyozkin: @MikeEdgar 
We'll need to deprecate then the group and sub path setters/getters in `JWTAuthContextInfo` and return the arrays and 'lose' the complete path for the log purposes when a given token does not match etc and I'm not 100% sure it is worth it, as it is likely a good number of the custom paths just point to a top level claim which is different to `groups` and such we'll be optimizing for like 20%. What do you think about a minor optimization for this what appears to be the typical case: split only if a '/' character is contained - it would allow us just to get the claim path value as is in 80% ?
Thanks

MikeEdgar: I agree that most apps will use this to identify another top-level claim. Checking for the `/` delimiter first is a good idea.

sberyozkin: Sounds good. So indeed, the custom claims case in itself is probably 30% max, and within that percentage with this simple check we can limit the pattern split to a likely 5% or 10% max case. let me start with it and I'll have no problems with going the deprecation route later if there be some evidence the split is used quite a lot

sberyozkin:
@MikeEdgar Mike, have a look please if it is better now, I've also updated the docs to note that the double quotes have to be used with the namespace qualified claims names.

sberyozkin:
@MikeEdgar Thanks

syndic8-stephen:
@sberyozkin. Thanks, got pulled in another direction last minute for a bit and did not have time to get back to this.  Looks good.

PR #183, opened 2020-02-11
Add a key format property

sberyozkin:
`KeyFormat` property is introduced in preparation for #85 and also for `KeyLocationResolver` to avoid trying all the supported formats (in the worst cases, if it is HTTPS based Base64URL-encoded JWK set, the key will be loaded on the 5th attempt).

I've been thinking if `KeyStoreType` would be better given that it will be `JKS` which will be added next. But then for example an individual PEM key is not exactly a key store though it can be seen as such. So I've added a note in the docs it can be either the format or store type

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
@@ -50,6 +50,7 @@ import org.jose4j.keys.resolvers.VerificationKeyResolver;
 import org.jose4j.lang.JoseException;
 import org.jose4j.lang.UnresolvableKeyException;
 
+import io.smallrye.jwt.KeyFormat;
 import io.smallrye.jwt.KeyUtils;
 import io.smallrye.jwt.algorithm.SignatureAlgorithm;
 
@@ -162,7 +163,8 @@ public class KeyLocationResolver implements VerificationKeyResolver {
 
     protected void initializeKeyContent() throws Exception {
 
-        if (authContextInfo.getPublicKeyLocation() != null && authContextInfo.getPublicKeyLocation().startsWith(HTTPS_SCHEME)) {
+        if (mayBeFormat(KeyFormat.JWK) && authContextInfo.getPublicKeyLocation() != null
+                && authContextInfo.getPublicKeyLocation().startsWith(HTTPS_SCHEME)) {
             LOGGER.debugf("Trying to load the keys from the HTTPS JWK(S)...");
             httpsJwks = initializeHttpsJwks();
             httpsJwks.setDefaultCacheDuration(authContextInfo.getJwksRefreshInterval().longValue() * 60L);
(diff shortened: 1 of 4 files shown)


implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -243,6 +245,14 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
     @ConfigProperty(name = "smallrye.jwt.verify.algorithm", defaultValue = "RS256")
     private Optional<SignatureAlgorithm> signatureAlgorithm;
 
+    /**
+     * Supported key format. By default a key can be in any of the supported formats:
+     * PEM key, PEM certificate, JWK key set or single JWK (possibly Base64URL-encoded).
+     */
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.verify.key-format", defaultValue = "ANY")
+    private Optional<KeyFormat> keyFormat;
```
MikeEdgar: What are your thoughts on removing the `Optional`? Since a `defaultValue` is given, I'm not sure it adds anything. I know this is not the only property like this, so it's not unusual. From the standpoint of repetition, the default is given three times - here, in the `orElse` and also as the default in `JWTAuthContextInfo`.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -178,16 +180,28 @@ protected void initializeKeyContent() throws Exception {
...
         // Try to init the verification key from the local PEM or JWK(S) content
-        verificationKey = tryAsPEMPublicKey(content, authContextInfo.getSignatureAlgorithm());
-        if (verificationKey == null) {
+        verificationKey = null;
+        if (mayBePemKey()) {
+            verificationKey = tryAsPEMPublicKey(content, authContextInfo.getSignatureAlgorithm());
+            if (isPemKey()) {
+                return;
+            }
+        }
```
MikeEdgar: I'm wondering if the check for `isPemKey` should instead check for `verificationKey !=null`. Suppose `KeyFormat.ANY` and a PEM key is successfully parsed. Same goes for the other types below.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -349,4 +363,36 @@ public InputStream resolve(String keyLocation) throws IOException {
...
         }
     }
+
+    boolean mayBePemKey() {
+        return isPemKey() || authContextInfo.getKeyFormat() == KeyFormat.ANY;
+    }
+
+    boolean isPemKey() {
+        return authContextInfo.getKeyFormat() == KeyFormat.PEM_KEY;
+    }
```
MikeEdgar: What do you think about reducing the number of methods by using a parameter? E.g.

```java
boolean mayBeFormat(KeyFormat format) {
    return isFormat(format) || authContextInfo.getKeyFormat() == KeyFormat.ANY;
}

boolean isFormat(KeyFormat format) {
    return authContextInfo.getKeyFormat() == format;
}
```

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: @MikeEdgar That was a polite way of suggesting the big redundancy is here, thanks :-). Yes, I keep copying and pasting. I'll start from fixing it from this PR

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: @MikeEdgar In that case the existing `if (verificationKey == null ...)` guards will prevent any other attempts to load the key. And we also don't want the check to continue if `verificationKey` is null but the user asked for `PEM_KEY`. 
I can remove the existing `verificationKey == null ` guards and instead have
```
if (verificationKey != null || isPemKey()) {
return;
}
```
and do it within every other if block but I'm not sure it will make things clearer. What is your preference ?

MikeEdgar: @sberyozkin - I got it now. I say keep it as-is.

sberyozkin:
@MikeEdgar Hi, sorry for a delay, got sidetracked a bit. Have a look please at the latest updates, I thought I'd also move the verification key checks inside individual loops as per your original idea, it looks even cleaner in the IDE, and a minor saving too, no extra checks for the successful PEM load, etc, thanks

sberyozkin:
@MikeEdgar Thanks

Issue #188, opened 2020-02-26
Cookie name property should default to MPJWT


sberyozkin:
The spec recommends recognizing `MPJWT` name. Thus if the header property is set to `Cookie` then `MPJWT` cookie will be checked by default

sberyozkin:
Right now it is defaulted to `Bearer`. We need to decide if we should keep it or use an opportunity (move to `2.1.0`) and replace with `MPJWT` or simply check both `Bearer` and `MPJWT` cookie names :-). The last or 2nd options look good to me.
CC @MikeEdgar @darranl

MikeEdgar:
@sberyozkin - which version of the spec are you referring to? I didn't see that recommendation in the current PDF or the asciidoc on `master`.

sberyozkin:
@MikeEdgar Oh, I was looking at Scott's PR. The section related to the cookies has been reverted by David:
https://github.com/eclipse/microprofile-jwt-auth/commit/96cfcd0bd4b4a280388a1e0e20987965a654c708

sberyozkin:
Let me close this one for now

PR #189, opened 2020-02-26
Add configuration to always check `Authorization` header

nicolaimainiero:
- add new property `smallrye.jwt.always-check-authorization` with default false
- evaluate `smallrye.jwt.always-check-authorization` in `AbstractBearerTokenExtractor` if set to true, fallback to `Authorization` header even if `tokenHeaderName` is set to `Cookie` but cookie is empty
- Add smallrye.jwt.always-check-authorization` in documentation with description

As discussed with @sberyozkin on zulipchat

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -45,6 +45,7 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
 |smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
 |smallrye.jwt.token.cookie|none|Name of the cookie containing a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
+|smallrye.jwt.always-check-authorization|false|If `true` the `Authorization` header is used if the cookie doesn't contain a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
 |smallrye.jwt.token.schemes|`Bearer`|Comma-separated list containing an alternative single or multiple schemes, for example, `DPoP`.
 |smallrye.jwt.token.kid|none|Key identifier. If it is set then the verification JWK key as well every JWT token must have a matching `kid` header.
 |smallrye.jwt.time-to-live|none|The maximum number of seconds that a JWT may be issued for use. Effectively, the difference between the expiration date of the JWT and the issued at date must not exceed this value.
--- a/README.adoc
+++ b/README.adoc
@@ -45,7 +45,7 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
 |smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
 |smallrye.jwt.token.cookie|none|Name of the cookie containing a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
-|smallrye.jwt.always-check-authorization|false|If `true` the `Authorization` header is used if the cookie doesn't contain a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
+|smallrye.jwt.always-check-authorization|false|Set this property to `true` for `Authorization` header be checked even if the `smallrye.jwt.token.header` is set to `Cookie` but no cookie with a `smallrye.jwt.token.cookie` name exists.
 |smallrye.jwt.token.schemes|`Bearer`|Comma-separated list containing an alternative single or multiple schemes, for example, `DPoP`.
 |smallrye.jwt.token.kid|none|Key identifier. If it is set then the verification JWK key as well every JWT token must have a matching `kid` header.
 |smallrye.jwt.time-to-live|none|The maximum number of seconds that a JWT may be issued for use. Effectively, the difference between the expiration date of the JWT and the issued at date must not exceed this value.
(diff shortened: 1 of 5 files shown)


README.adoc
```diff
@@ -45,6 +45,7 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
 |smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
 |smallrye.jwt.token.cookie|none|Name of the cookie containing a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
+|smallrye.jwt.always-check-authorization|false|If `true` the `Authorization` header is used if the cookie doesn't contain a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
```
sberyozkin: I'd like to propose to rephrase (or may be simplify) it a bit:
```
Set this property to `true` for `Authorization` header be checked even if the `smallrye.jwt.token.header` is set to `Cookie` but no cookie with a `smallrye.jwt.token.cookie` name exists.
```

implementation/src/main/java/io/smallrye/jwt/auth/principal/JWTAuthContextInfo.java
```diff
@@ -37,6 +37,7 @@
     private Integer jwksRefreshInterval;
     private String tokenHeader = "Authorization";
     private String tokenCookie;
+    private boolean alwaysCheckAutorization;
```
sberyozkin: Minor typo in the name

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -305,6 +313,8 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
             contextInfo.setTokenHeader(tokenHeader);
         }
 
+        contextInfo.setAlwaysCheckAutorization(alwaysCheckAuthorization);
+
```
sberyozkin: This class also has a copy constructor and the property getters in the end, please update those

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -141,6 +141,14 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
     @ConfigProperty(name = "smallrye.jwt.token.cookie")
     private Optional<String> tokenCookie;
 
+    /**
+     * Should check `Authorization` header if cookie didn't contain a token. This property is ignored unless the
```
sberyozkin: Similarly to the earlier comment, lets rephrase to
```
If `true` then the `Authorization` header will be checked even if the `smallrye.jwt.token.header` is set to `Cookie` but no cookie with a `smallrye.jwt.token.cookie` name exists.
```

sberyozkin:
Hi @nicolaimainiero it looks very good, thanks, few cosmetic changes are proposed.

@MikeEdgar, @darranl if you'd like to propose some better wording for the docs for this new property then please don't hesitate to suggest them :-). The use case is to support both `Authorization` and cookies at the same time to deal with different type of clients as described by @nicolaimainiero in one of Quarkus Zulip threads

sberyozkin:
@nicolaimainiero By the way, while you are dealing with the cookie related code :-), would you mind doing another minor update related to #188 ? Just set a default `smallrye.jwt.token.cookie` value to `MPJWT` and remove `Optional`, we'll be moving to `2.1.0` so it will not a problem. Thanks

nicolaimainiero:
Sure, I can do #188 . So if I understand this correct, I can also remove the `setContextTokenCookie()` in `SmallryeJwtUtils` since I define the defauklt value of the cookie through the annotation in `JWTAuthContextInfoProvider`. Should the change be made like this?

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
nicolaimainiero: I updated the property getters, but what do you mean by copy constructor.

sberyozkin: @nicolaimainiero sorry, [meant this factory method](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java#L73).

sberyozkin:
@nicolaimainiero Thanks, I'm having a bit of doubt now about this `MPJWT` cookie name being a default value of `smallrye.jwt.cookie.name` since it makes sense only if the header is set to `Cookie`. Lets simplify it, please update the abstract code you have already updated in this PR,  use `MPJWT` there as a cookie name if no cookie name has been set in `JwtAuthContextInfo`.
Thanks

sberyozkin:
@nicolaimainiero I'm sorry, lets ignore this `MPJWT` cookie name idea altogether for the moment, my fault I haven't seen it already has a default value `Bearer`, I will discuss it with @MikeEdgar if he is OK to replacing it with `MPJWT` at #188, will also need to check with Darran.
So yes, please only address few minor comments related to this PR and it will be good to go IMHO, thanks

nicolaimainiero:
Hi @sberyozkin, i've updated the PR with the requested changes regarding typos and wording.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -79,6 +79,7 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
         provider.mpJwtRequireIss = Optional.of(Boolean.TRUE);
         provider.tokenHeader = AUTHORIZATION_HEADER;
         provider.tokenCookie = Optional.empty();
+        provider.alwaysCheckAuthorization = false;
```
sberyozkin: @nicolaimainiero Can you drop this one please, I know I've asked you to do it but I forgot the default value was 'false', my apologies

sberyozkin:
Hi @nicolaimainiero Thanks, one last update please, see above, my fault really...

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
nicolaimainiero: Sure. I was unsure if it was necessary because some properties are configured as default values and some in this method. Nevertheless, I have now removed it.

PR #190, opened 2020-02-27
Move version to 2.1

darranl:
We have been discussing this week the need to be able to tag maintenance releases that contain bug fixes only without new enhancements, our integration within WildFly is now in the closing stages before we tag we should move to the next minor in case micro bug fix releases are needed.


sberyozkin:
@darranl It makes sense. I suppose, once `MP JWT 1.2` is out we will move to `3.0.0`, and then eventually to `4.0.0` once MP JWT 2.0 (MP 4.0) is out... thanks

PR #194, opened 2020-02-28
Make JWTCallerPrincipalFactory injectable

### LINKED ISSUE #193: Why is JWTCallerPrincipalFactory not injected into DefaultJWTParser
In `DefaultJWTParser`, one field (`JWTAuthContextInfo authContextInfo`) is injected, but the other (`JWTCallerPrincipalFactory callerPrincipalFactory`) isn't. The `callerPrincipalFactory` is generated via reflection code in the `JWTCallerPrincipalFactory.instance()` method (see https://github.com/smallrye/smallrye-jwt/blob/08970ef586acb9414f0ebfcdb3fc8566e28b22cc/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java#L46).

If there is no special reason for this behavior, I think both should be injected the same way. The current implementation makes it hard to switch out the implementation with an own implementation which requires injection.

andreas-eberle:
This PR allows injecting own implementatons of the `JWTCallerPrincipalFactory` into the `DefaultJWTParser`.

Edit: I don't think the `JWTCallerPrincipalFactory.instance()` method should be removed, since it allows a simple entrance into the library. However, since the `DefaultJWTParser` already uses injection, both dependencies should be injected the same way for easier testability and configurability.

Fixes #193.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTCallerPrincipalFactory.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTCallerPrincipalFactory.java
@@ -16,21 +16,22 @@
  */
 package io.smallrye.jwt.auth.principal;
 
+import javax.enterprise.context.ApplicationScoped;
+
 import org.jose4j.jwt.consumer.JwtContext;
 
 /**
  * A default implementation of the abstract JWTCallerPrincipalFactory that uses the Keycloak token parsing classes.
  */
+@ApplicationScoped
 public class DefaultJWTCallerPrincipalFactory extends JWTCallerPrincipalFactory {
 
     private final DefaultJWTTokenParser parser = new DefaultJWTTokenParser();
 
     @Override
     public JWTCallerPrincipal parse(final String token, final JWTAuthContextInfo authContextInfo) throws ParseException {
-
         JwtContext jwtContext = parser.parse(token, authContextInfo);
         String type = jwtContext.getJoseObjects().get(0).getHeader("typ");
         return new DefaultJWTCallerPrincipal(type, jwtContext.getJwtClaims());
     }
-
 }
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTCallerPrincipalFactory.java
```diff
@@ -16,11 +16,14 @@
...
 package io.smallrye.jwt.auth.principal;
 
+import javax.enterprise.context.ApplicationScoped;
+
 import org.jose4j.jwt.consumer.JwtContext;
 
 /**
  * A default implementation of the abstract JWTCallerPrincipalFactory that uses the Keycloak token parsing classes.
  */
+@ApplicationScoped
```
sberyozkin, line 26: @andreas-eberle TCK is failing, perhaps there should be `JWTCallerPrincipalFactoryProducer` which will use the method you removed in `DefaultJWTParser` to produce a default factory, and the custom factories will be injected with the producers with `@Alternative` ?
Please also rebase
Thanks

andreas-eberle: I was able to fix it by using the same mechanism used for DefaultJwtParser in `SmallRyeJWTAuthCDIExtension.java`.

andreas-eberle:
Rebased and tests seem to work.

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTCallerPrincipalFactory.java
sberyozkin: @andreas-eberle thanks

sberyozkin:
@andreas-eberle Can you please give me a favour and run it against Quarkus master ?
CC @darranl Hi Darran, can you please try this PR against WildFly if there is some concern ?

sberyozkin:
@andreas-eberle I'll test with Quarkus, can you please add a test, probably in `testsuite/basic`, showing for example, a custom factory which skips the verification completely, to confirm that those cases where JWT is verified earlier can now be supported ?

andreas-eberle:
@sberyozkin: I'm a bit constraint with my time at the moment. I'll try to get back to it as soon as possible.

sberyozkin:
Fixed as part of #298

sberyozkin:
Fixed with #298, thanks

PR #198, opened 2020-03-14
Remove redundant io.smallrye.config property

sberyozkin:


sberyozkin:
This is a redundant test property so I will just go ahead and merge

PR #200, opened 2020-03-19
Add TCK tests to project.

radcortez:
Here is the initial work to run the TCK directly in the project.

Still some tests failures, probably related with missing setup that need to be debugged on a case by case basis.

This is using `io.smallrye.jwt.auth.jaxrs.JWTAuthenticationFilter` to apply the security context.

A few considerations:
- It seems the filters are not used either in Elytron (uses their internal HttpServerAuthenticationMechanism) or Quarkus (they implement their own with Vert.x).
- The filters were not being testes previously, so some of the failures related with the filters I'm not sure if they are cause by the setup or some bug in the filter.
- Do we want to keep the filters and change them is required? Is anyone else using them?
- The `SmallRyeJWTAuthCDIExtension` has Optional injection turned off by default, but this is required to pass the TCK. I'm wondering if this was something between 1.0 and 1.1? Elytron extends the extension and turn that on. I think we should just turn that on by default.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,7 @@
     <version.io.smallrye.config>1.7.0</version.io.smallrye.config>
     <version.microprofile.config>1.4</version.microprofile.config>
     <version.jose4j>0.7.0</version.jose4j>
+    <version.resteasy>3.9.1.Final</version.resteasy>
     <version.mokito>3.3.3</version.mokito>
     <version.weld-junit4>2.0.1.Final</version.weld-junit4>
     <sonar.coverage.jacoco.xmlReportPaths>${project.basedir}/testsuite/basic/target/site/jacoco-aggregate/jacoco.xml,${project.basedir}/../testsuite/basic/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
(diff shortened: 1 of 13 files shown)


radcortez:
By the way, don't merge this yet. I should have created this as draft.

radcortez:
@sberyozkin let me know what you think about the comments.

radcortez:
Ok, I've made some additional fixes on the setup that shouldn't need to mess with the filters and all the TCK is passing now :)

sberyozkin:
@radcortez Wow, super fast, really is, nice work, thank you :-)

radcortez:
I did copy some stuff from other projects :)

sberyozkin:
@radcortez Hi Roberto, is it possible to get that `test-basic-suite.xml` extracted from the tck jar instead with the dependency plugin or is it too messy ?

radcortez:
I think it should be possible. Let me have a look.

radcortez:
@sberyozkin ops. It is not going to work right now. The TCK jars are not including the suite files. In fact they are just in the root of the project and the README just states to copy them.

I think we can chance it and start including those files. I'll open a PR in MP JWT for it.

sberyozkin:
@radcortez Sounds good, in meantime let me approve this PR, cheers

PR #203, opened 2020-03-20
KeyLocationResolver, give debug information about authContextInfo

SchulteMarkus:
Recently I had to debug my usage of smallrye-jwt (using a local jwks.json). This PR tries to improve debugging in this area:
- First of all, log-debug AuthContextInfo in KeyLocationResolver
- 2nd, no more duplicate "Failed to load a key from [...]" log-debug message

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
@@ -80,11 +80,13 @@ public class KeyLocationResolver implements VerificationKeyResolver {
 
     public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws UnresolvableKeyException {
         this.authContextInfo = authContextInfo;
+        LOGGER.debugf("AuthContextInfo is: %s", authContextInfo);
+
         try {
             initializeKeyContent();
         } catch (Exception e) {
             throw new UnresolvableKeyException("Failed to load a key from "
-                    + (authContextInfo.getPublicKeyContent() != null ? " the 'mp.jwt.verify.publickey' property"
+                    + (authContextInfo.getPublicKeyContent() != null ? "the 'mp.jwt.verify.publickey' property"
                             : authContextInfo.getPublicKeyLocation()),
                     e);
         }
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -79,12 +79,18 @@
     private JWTAuthContextInfo authContextInfo;
 
     public KeyLocationResolver(JWTAuthContextInfo authContextInfo) throws UnresolvableKeyException {
+        if (authContextInfo == null) {
```
sberyozkin: This can not happen, so can be dropped IMHO. And if NPE did happen later then it would indicate a really broken default implementation :-)

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -79,12 +79,18 @@
...
+            throw new UnresolvableKeyException("No AuthContextInfo given");
+        }
+
         this.authContextInfo = authContextInfo;
+        LOGGER.debugf("AuthContextInfo is: %s", authContextInfo);
```
sberyozkin: This will call `toString` even if `debug` is not enabled ? I propose to check if debug is enabled and then log to avoid the redundant String allocation

darranl: This should be Ok, this is the implementation of debugf: -

    public void debugf(String format, Object param1) {
        if (isEnabled(Level.DEBUG)) {
            doLogf(Level.DEBUG, FQCN, format, new Object[] { param1 }, null);
        }
    }

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
SchulteMarkus: Removed the check.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: @darranl Hi Darran, this is fine, I wonder, will the compiled code will effectively have something like:
```
LOGGER.debugf("AuthContextInfo is: %s", authContextInfo.toString());
```
?

SchulteMarkus: In my experience would be quite unusual to encapsulate this by ```LOGGER.isDebugEnabled()```, but of course can do so, if you wish.

SchulteMarkus: > @darranl Hi Darran, this is fine, I wonder, will the compiled code effectively have something like:
> 
> ```
> LOGGER.debugf("AuthContextInfo is: %s", authContextInfo.toString());
> ```
> 
> ?

From .class file
```java
LOGGER.debugf("AuthContextInfo is: %s", authContextInfo);
```

sberyozkin: @SchulteMarkus, thanks for checking, then obviously we should keep your original code :-), Darran, @darranl, thanks as well

PR #206, opened 2020-03-25
Support Optional JsonWebToken injection.

radcortez:
Prototype work to support https://github.com/eclipse/microprofile-jwt-auth/issues/120


sberyozkin:
@radcortez In that case please mark is as a Draft PR (and it should be in `testsuite/basic`), thanks

radcortez:
Yeah, its a pain because if you don't open it as a Draft, there is no way to change it and I forget when I've opened. I guess I can close this one and open a new one.

I've added it in the TCK, because since this was some exploratory work for eclipse/microprofile-jwt-auth#120, it would then be easier to just donate the test we wrote to the official TCK.

Issue #80, opened 2019-06-19
Introduce a property for listing the required claims


sberyozkin:
For example, having `smallrye.jwt.required.claims`=`a,b,c` would fail the verification if at least one of these claims is missing

radcortez:
I'm going to pick this up.

PR #209, opened 2020-03-26
#80 - Added configuration to validate required claims in the JWT.

radcortez:


sberyozkin:
Hi @radcortez Thanks for picking it up. I've been contemplating recently about this issue, since some claims such as the `issuer` and `audiences` can already be enforced, `require-named-principal` indirectly enforces too. 
But having a generic mechanism is still useful IMHO, to cover all sort of claims, so that the users can write the safe code without null checks... So yeah, it is good to have it :-)

radcortez:
Cool :)

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -138,6 +140,22 @@ void setExpectedAudience(JwtConsumerBuilder builder, JWTAuthContextInfo authCont
         }
     }
 
+    private void setRequiredClaims(JwtConsumerBuilder builder, JWTAuthContextInfo authContextInfo) {
```
sberyozkin: @radcortez We've discussed with @darranl before, and with Stuart in Quarkus, but in general we just prefer a simple old Java code which performs better. 
But in this case IMHO we don't need registering the custom verificators at all.
[Over here](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java#L93), you can add a simple function passing the parsed claims and context, `JwtClaims` has a `getClaimsAsMap` or similar, then the following one liner will do
```
if (!claims.getAsMap().keySet().containsAll(requiredClaims)) {
    throw new InvalidJwtException("Required claims are not available");
}
``` 

Something like that :-)

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -273,6 +274,14 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
     @ConfigProperty(name = "smallrye.jwt.verify.aud")
     Optional<Set<String>> expectedAudience;
 
+    /**
+     * List of claim names that must be present in the JWT for it to be valid. The configuration should be specified
+     * as a comma-separated list per MP Config requirements for a collection property.
```
sberyozkin: @radcortez Lets drop ` per MP Config requirements for a collection property` - we already have few properties like this one where we don't justify why (earlier Set would not even work). And please update README with the same description

sberyozkin:
@radcortez looks good, thanks. I'd like to CC to @MikeEdgar - this issue was opened after our conversation as far as I recall. Thanks

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -265,6 +267,21 @@ private void verifyTimeToLive(JWTAuthContextInfo authContextInfo, JwtClaims clai
         }
     }
 
+    private void verifyRequiredClaims(JWTAuthContextInfo authContextInfo, JwtClaims claimsSet) throws ParseException {
+        final Set<String> requiredClaims = authContextInfo.getRequiredClaims();
+
+        if (requiredClaims != null) {
+            final String missingClaims = requiredClaims.stream()
```
sberyozkin: H @radcortez Nearly there :-), as I implied, lets avoid the stream API. I still think the simple keyset check would suffice.
If the check fails then there must be a `Collection` function which would return the keys which are not intersecting - though  we don't really need to report back what exactly is required - this is usually avoided in the security flows to keep the attackers guessing.

 
It should be `InvalidJwtException` internally - because it is invalid in this case, and then it will be wrapped into a `ParseException` and propagated.

README.adoc
```diff
@@ -58,6 +58,7 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |smallrye.jwt.jwks.refresh-interval|60|JWK cache refresh interval in minutes. It will be ignored unless the `mp.jwt.verify.publickey.location` points to the HTTPS URL based JWK set and no HTTP `Cache-Control` response header with a positive `max-age` parameter value is returned from a JWK HTTPS endpoint.
 |smallrye.jwt.expiration.grace|60|Expiration grace in seconds. By default an expired token will still be accepted if the current time is no more than 1 min after the token expiry time.
 |smallrye.jwt.verify.aud|none|Comma separated list of the audiences that a token `aud` claim may contain.
+|smallrye.jwt.required.claims|none|Comma separated list of required claims for the token to be valid.
```
sberyozkin: @radcortez May be worth syncing it here with the `aud` text ?
```
Comma separated list of the required claims that a token must contain.
```
It takes a few small iterations to settle on the text :-), I'd not likely propose this last update if the `aud` text were not just above it :-)

sberyozkin: In fact even the `required` word can be dropped:
```
Comma separated list of the claims that a token must contain.
```
`must contain` implies required

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
radcortez: Hum, but doing a `keySet().containsAll` and then another `Collection` function to retrieve the keys that not intersect, isn't more inefficient than using the `Stream` stream does everything in a single pass of the date, while in here we need to perform at least 2 iterations.

Plus, I've seen the use of the `Stream` API in Quarkus before :)

sberyozkin: @radcortez That another `Collection` function can be dropped, as I said reporting what is required back is not needed. Or if you'd really like, if the key set checks fails and debug mode is enabled, run the extra collection function there and log what is missing but still report the opaque enough message back :-)
I'm sure Quarkus has plenty of this code but in this case it is a runtime time, smallrye-jwt is also used in WildFly among other containers, and Darran has already provided one PR to get rid of some stream api so lets continue in this spirit
thanks

radcortez: Ok. Thanks. Just wanted to clarify the reason.

I think we should always have this information for debug purposes, or it is usually a pain to figure out why a token is valid / invalid. (Ops someone changed the configurations and we are now requiring claim X :) )

radcortez:
OK, new push :)

sberyozkin:
Hi Roberto @radcortez sorry for a delay, thanks, the the filter API, it is probably Ok to have it in the debug mode (which I implied earlier so that your filter code could be kept :-) ) as I guess the extra performance savings are of the lesser priority when someone is running in the debug mode, but lets keep the current code anyway now.
The only thing which is still needed is to remove a duplicate dot in the README at the end of the new property description :-), I promise it is the last review comment from me :-)

radcortez:
No worries :)

Let me readdd the Stream filter then because I hate that code :)

radcortez:
Ok. Done!

sberyozkin:
@radcortez Thanks

PR #210, opened 2020-03-26
Support Optional JsonWebToken injection.

radcortez:
Original PR here: https://github.com/smallrye/smallrye-jwt/pull/206

This PR will depend of the consensus reached in https://github.com/eclipse/microprofile-jwt-auth/issues/120.


sberyozkin:
@radcortez I believe this one can be closed

radcortez:
We can for now. We may need to reuse some stuff depending on what happens with https://github.com/eclipse/microprofile-jwt-auth/issues/120.

PR #213, opened 2020-03-31
Formatter Maven Plugin changes.

radcortez:


darranl:
If there is any way within CI it would help if PRs could fail if the build results in code being reformatted to stop these slipping through.

radcortez:
I think so. We can run this goal instead in the CI:
https://code.revelc.net/formatter-maven-plugin/validate-mojo.html

And fail the build. I'll give it a try.

radcortez:
I'll be merging this, because I'm getting annoyed of file changes not related with the work I'm doing :)

radcortez:
@darranl The check to fail the CI if the formatter fails is already in the CI as well, so hopefully this shouldn't happen anymore.

PR #214, opened 2020-03-31
Fail the CI build if formatter rules fail.

radcortez:


radcortez:
Fail due to formatter code rules. Fix in #213.

PR #219, opened 2020-04-13
Fixes #216 - Move current documentation to Antora.

### LINKED ISSUE #216: Move documentation to the Antora structure
This will allow for the documentation to be published in SmallRye website.

radcortez:
This is to align with the SmallRye Documentation Process: https://github.com/smallrye/smallrye-parent/wiki/SmallRye-documentation-process.

I've just moved the current documentation to the new structure. If we are happy with this, we can later remove the old structure and simplify the main README.


doc/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,16 @@
...
+Currently it is focused on supporting the MP-JWT 1.1 spec, and primarily deals with the parsing of the JWT string into
+a JsonWebToken implementation.
+
+In the future, when MP-JWT 2.0 can build on the JSR-375 security APIs, there should be more support for defining the CDI extension and security layer integration.
+
+== Runtimes
+
+Runtimes that currently use it as their JWT RBAC implementation include:
+
+* https://quarkus.io/[Quarkus]
```
sberyozkin, line 14: @radcortez Is this section recommended across all of Smallrye ? @darranl Are you OK with WildFly referenced ?

doc/modules/ROOT/pages/parsing.adoc
```diff
@@ -0,0 +1,37 @@
+:doctype: book
```
sberyozkin, line 1: @radcortez IMHO we need to drop this page, it provides some internal details which we should be free to change.
Perhaps a short intro to `JsonWebToken` with the class showing it injected can be added instead (example, simple service which returns a principal name by calling `jwt.getName()`)

sberyozkin:
@radcortez Thanks for doing this transition, it will surely look much nicer, minor updates suggested, cheers

doc/modules/ROOT/pages/index.adoc
radcortez: Other Smallrye projects documentation do this reference. I see no harm in having it, but we can remove it.

doc/modules/ROOT/pages/parsing.adoc
radcortez: I copied this from the existent documentation:
https://github.com/smallrye/smallrye-jwt/blob/6390becaa77ece913a365fcaf05c8d469caf0dfd/docs/src/main/asciidoc/index.adoc

Since this is specific SmallRye documentation, I see no harm in providing documentation on the internal, even if they change later (as long as we update the documentation :)).

doc/modules/ROOT/pages/index.adoc
sberyozkin: @radcortez Sure it is a good idea in general; just would like a confirmation he is fine with WildFly being referenced

doc/modules/ROOT/pages/parsing.adoc
sberyozkin: @radcortez I did not see it, it is an outdated doc, https://github.com/smallrye/smallrye-jwt lists those properties, and `requireiss` was left out intentionally. That parsing info is irrelevant to the users who work with JsonWebToken soi needs to be dropped IMHO

sberyozkin:
Hi @radcortez Can you please give me a favour and copy [this doc](https://quarkus.io/guides/security-jwt#generate-jwt-tokens) here too ? That would be more relevant for sure than the internal info about parsing the token :-) Cheers

radcortez:
> Hi @radcortez Can you please give me a favour and copy [this doc](https://quarkus.io/guides/security-jwt#generate-jwt-tokens) here too ? That would be more relevant for sure than the internal info about parsing the token :-) Cheers

Sure.

doc/modules/ROOT/pages/parsing.adoc
radcortez: Very well, I'll remove it.

sberyozkin: @radcortez since then a parser interface has also been introduced, so at the moment it is a bit of mute point as to how the implementations integrate it. My understanding that none of the integrators are  dealing directly with it like this, Thorntail has a custom extension but as far as I recall it just depends on the producer to have `JsonWebToken` injected. Perhaps later on we can come up with the integration advice section and list few ideas. Lets review it later, perhaps some of this parse text can return, please open an issue so that we can discuss the ideas there. Thanks

radcortez:
Ok done. Just need the confirmation to include Wildfly.

PR #226, opened 2020-04-22
Fix #222 - Consistent injection and conversion on all supported types of Claims.

### LINKED ISSUE #222: Consistent Injection combinations of Claims
There are multiple ways to inject values from a `JWT`. From the latest MP JWT Spec, the following injection combinations must be supported (using `Long` as an example):

```java
@Inject
@Claim("long")
private long longClaim;
@Inject
@Claim("long")
private Long longClaimWrapper;
@Inject
@Claim("long")
private ClaimValue<Long> longClaimValue;
@Inject
@Claim("long")
private Optional<Long> longOptional;
@Inject
@Claim("long")
private ClaimValue<Optional<Long>> longClaimValueOptional;
@Inject
@Claim("long")
private JsonNumber longJson;
@Inject
@Claim("long")
private Optional<JsonNumber> longOptionalJson;
@Inject
@Claim("long")
private ClaimValue<JsonNumber> longClaimValueJson;
@Inject
@Claim("long")
private ClaimValue<Optional<JsonNumber>> longClaimValueOptionalJson;
```

There is also valid for other types like `Boolean`, `String` and corresponding `JsonValue` types. In #218, we realized that SmallRye JWT is not covering all cases and should be more consistent in how injection is done for the multiple combinations. The current implementation is able to handle the following injection combinations:

| Injection Type | Result |
| -------------  | -----|
|`long` | OK |
|`Long`| OK |
|`ClaimValue<Long>`| Fails |
|`Optional<Long>`| Fails |
|`ClaimValue<Optional<Long>>`| Fails | 
|`JsonNumber`| OK |
|`Optional<JsonNumber>`| OK |
|`ClaimValue<JsonNumber>`| OK |
|`ClaimValue<Optional<JsonNumber>>`| OK |

The main reason for injection to fail is a `ClassCastException` because values in the `claimSet` are usually stored as a `JsonValue` (except if it is a known claim and the type is set in the `Claims` enum).

So we need to determine the right type at injection point and then provide a conversion if required so the types match and injection work.

radcortez:


implementation/src/test/java/io/smallrye/jwt/auth/cdi/ClaimInjectionTest.java
```diff
@@ -21,34 +32,121 @@
...
-    public void inject() {
+    public void injectBoolean() {
         assertTrue(claimsBean.isBooleanClaim());
         assertTrue(claimsBean.getBooleanClaimWrapper());
-        //assertTrue(claimsBean.getBooleanClaimValue().getValue()); // does not unwrap json to wrapper type
+        assertTrue(claimsBean.getBooleanClaimValue().getValue());
+        assertTrue(claimsBean.getBooleanOptional().get());
+        assertTrue(claimsBean.getBooleanClaimValueOptional().getValue().get());
+        // This doesn't work because it causes ambiguous dependencies for JsonValue in JsonValueProducer
+        //assertEquals(claimsBean.getBooleanJson(), JsonValue.TRUE);
```
sberyozkin, line 68: @radcortez Can you clarify please ?

sberyozkin:
@radcortez Please check it against Quarkus as some code is done there to support JSON values in quarkus-smallrye-jwt

implementation/src/test/java/io/smallrye/jwt/auth/cdi/ClaimInjectionTest.java
radcortez: Unfortunately, this PR doesn't solve all the cases yet (it does solve most of them). The issue here is with injection of `JsonValue` to represent a `Boolean`. If you look into `JsonValueProducer` we have specialised producers for all Json types, but not for `JsonValue` because it is the super class for all Json types. 

If I add a `JsonValue` producer to solve the `Boolean` case, then we have ambiguous dependencies with all the other types, because `JsonValue` can also be resolver for all the other types.

To fix this case, I need to rewrite `JsonValueProducer` a bit and I wanted to avoid that in this PR.

radcortez:
> @radcortez Please check it against Quarkus as some code is done there to support JSON values in quarkus-smallrye-jwt

I did. It passes the current TCK (1.1) and the new one (1.2-SNAPSHOT). I've run both locally. I would prefer the CI to do it, but we don't publish SNAPSHOTS, so we can't do it.

sberyozkin:
@radcortez The new TCK includes the encryption tests (I have the code on my branch), did you exclude them ?

implementation/src/test/java/io/smallrye/jwt/auth/cdi/ClaimInjectionTest.java
sberyozkin: > To fix this case, I need to rewrite JsonValueProducer a bit and I wanted to avoid that in this PR.

+1 - lets iterate

radcortez:
I've used the current masters on each project. If the code is in your branch I didn't use it :)

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTCallerPrincipal.java
```diff
@@ -164,6 +164,10 @@ private void fixJoseTypes() {
                 replaceNumber(name);
             } else if (claimValue instanceof Boolean) {
                 replaceBoolean(name);
+            } else {
+                // Jose use plain Java types, and we convert stuff to Json. This was added to convert all the missing
+                // types, except String. Without this, we have some inconsistency (apart from String).
+                replaceObject(name);
```
sberyozkin, line 170: @radcortez What other missing types (apart from String) you refer to here ? `Map` covers a complex claim like `Address`, `List` - arrays (of whatever), `Number` - primitive and their wrappers (long/Long, etc), Boolean - booleans.

radcortez: I think I've added this in my previous code where I didn't have the `Boolean` method.

If we look closely, every case can be served by the `replaceObject`, because `JsonUtils.wrapValue` takes an `Object` and returns a `JsonValue`. There is not much point in casting the original claim type or the specific json type. I would propose to drop all `replaceX` methods and keep only `replaceObject`. What do you thin?

sberyozkin: @radcortez Good point, may be rename `replaceObject` as `replaceClaimValueWithJsonValue` or something like that, I though initially you meant replacing a map inside `replaceObject`. Or may be simply paste the content of `replaceObject` inside the `for loop` iterating over the custom claims, whatever you choose will be fine

(JsonUtils will need some further cleanup but lets deal with it in another issue - specifically arrays and maps will just have Java Integer/Long/etc) so it is only the top level long/etc custom claims are converted to JsonValue. But lets review it again later, to be honest I'd just return simple claim values as is (we already do it for String everywhere, so the same really makes sense for Long/etc) - it is a not a major issue though, we can introduce a property in case some users prefer dealing with JsonValue... But yeah, lets chat about it later

radcortez: Yes. I'll do that change and then we can figure something out with `JsonUtils`. One of the main issues is that our coverage is not great for this cases, so we do need to implement some tests to make sure we don't break.

sberyozkin:
@radcortez OK, in Quarkus the `tck-basic-suite.xml` is copied locally, so the last 5 tests here are missed, https://github.com/eclipse/microprofile-jwt-auth/blob/master/tck/src/test/resources/suites/tck-base-suite.xml#L78. But it is not a problem for now

PR #231, opened 2020-05-07
Force a JWK refresh if no JWK with a matching kid is available

### LINKED ISSUE #228: Refetch mp.jwt.verify.publickey.location when receiving unfamiliar kid
The OpenID Connect core specification says in section 10.1.1 that users of the jwks_uri should re-retrieve the key set when it is asked to verify a signature with a unknown kid. https://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys smallrye-jwt does not do this, leading to `smallrye.jwt.jwks.refresh-interval` seconds (default 60) of failed requests each time the openid connect server smallrye-jwt is used with rotates it's keys. 

For me there also seems to be some issue with the refresh-interval, where it some times doesn't even start working after the refresh interval, and the service needs to be restarted after a key rotation

sberyozkin:
Fixes #228 

This PR forces a refresh if no JWK with a matching kid is available. It also adds a check to avoid the frequent forced refreshes as recommended in a similar Quarkus discussion. (CC @stianst)

CC @sigmunau


implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
```diff
@@ -70,7 +77,28 @@ public void testLoadPemKeyWithWrongLocation() throws Exception {
...
+        lenient().doThrow(new RuntimeException()).when(httpsJwks).refresh();
+        keyLocationResolver = Mockito.spy(keyLocationResolver);
+        when(signature.getHeaders()).thenReturn(headers);
+        when(headers.getStringHeaderValue(JsonWebKey.KEY_ID_PARAMETER)).thenReturn("1");
+        assertEquals(key, keyLocationResolver.resolveKey(signature, emptyList()));
+        assertNull(keyLocationResolver.verificationKey);
+    }
+
+    @Test
+    public void testLoadHttpsJwksNonMathchingKid() throws Exception {
```
MikeEdgar, line 101: I'm a bit confused with this test. I don't think it is hitting the new code with the synchronization and call to `refresh`.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -128,18 +130,58 @@ private PublicKey tryAsJwk(JsonWebSignature jws) {
     }
 
     PublicKey getHttpsJwk(String kid) {
-        LOGGER.debugf("Trying to create a key from the HTTPS JWK(S)...");
+        try {
+            PublicKey theKey = getKeyFromJsonWebKeys(kid, httpsJwks.getJsonWebKeys(), authContextInfo.getSignatureAlgorithm());
+            if (theKey != null || isMatchingJwkAvailable(httpsJwks.getJsonWebKeys(), kid)) {
```
MikeEdgar, line 135: It may be possible for the two calls to `getJsonWebKeys` to return different lists depending on traffic/threads. It might be better to assign a local variable from a single call instead. If `theKey` is null and another thread refreshes the JWKS cache before the `isMatchingJwkAvailable`/`getJsonWebKeys`, I think it may return the null if the `kid` is found in the refreshed list. An edge case, but something to consider.

sberyozkin: Hi Mike, @MikeEdgar thanks, good catch, I had `List<JsonWebKey> keys = httpsJwks.getJsonWebKeys()` at some point but then I started optimizing :-) and when it is done prematurely, bugs do happen :-)

implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
sberyozkin: @MikeEdgar do you refer to  `lenient()` ? Without it, if I just have `doNothing()` Mockito complains.
In `testLoadHttpsJwksMathchingKid`, `RuntimeException` would be thrown if `refresh` is called.

My be here I need to get `RuntimeException` thrown if `refresh` is not called, no idea right now if it is possible but will check. thanks

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: Re the sync, I'd've gladly avoided it but I was explained it is a DOS risk otherwise, someone can keep messing with the kid header and if we do immediately every time it can seriously slow the system

sberyozkin: But this sync will be rarely hit, only briefly around the rotation period, so it is fine

MikeEdgar: @sberyozkin - that's what I figured as well. I looked over this section several times in conjunction with the implementation of `refresh` (which itself uses a semaphore), and I think the approach you took is good.

implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
MikeEdgar: @sberyozkin, the parts that are confusing me are the multiple calls to `setKeyId` with "1" and "2". I'm not clear that the call with the "2" has any effect. When I check the lines covered as reported by JaCoCo, it looks like the code to refresh to the token is not actually executed as I would expect based on the test method's name.

implementation/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
sberyozkin: Ok, thanks, resolving it now, a single call to get the list is made

implementation/src/test/java/io/smallrye/jwt/auth/principal/KeyLocationResolverTest.java
sberyozkin: @MikeEdgar I think I've fixed it and also added 1 more test to verify the refresh is declined if an attempt to made it too fast is done

MikeEdgar: @sberyozkin - looks good. I see the new process is running now.

sberyozkin:
@MikeEdgar Thanks

PR #232, opened 2020-05-08
Update some resource loading methods to check the context class loader

sberyozkin:
A simple update to check the context loader too as some resource loading methods do not work correctly in the Quarkus dev mode


sberyozkin:
@MikeEdgar Thanks

PR #234, opened 2020-05-11
Load the token resource from the file system

sberyozkin:
This PR addresses this [StackOverflow question](https://stackoverflow.com/questions/61713369/unable-to-create-instance-of-jwtclaimsbuilder-in-quarkus-application)

I'd like to release smallrye-jwt-2.1.2 once this PR is resolved

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/build/impl/JwtSigningUtils.java
+++ b/implementation/src/main/java/io/smallrye/jwt/build/impl/JwtSigningUtils.java
@@ -1,6 +1,8 @@
 package io.smallrye.jwt.build.impl;
 
 import java.io.BufferedReader;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -275,16 +277,29 @@ public class JwtSigningUtils {
     }
 
     static String readJsonContent(String jsonResName) {
-        try {
-            InputStream is = JwtSigningUtils.class.getResourceAsStream(jsonResName);
-            if (is == null) {
-                is = Thread.currentThread().getContextClassLoader().getResourceAsStream(jsonResName);
+        // Try as the class path resource
+        InputStream is = JwtSigningUtils.class.getResourceAsStream(jsonResName);
+        if (is == null) {
+            is = Thread.currentThread().getContextClassLoader().getResourceAsStream(jsonResName);
+        }
+        if (is == null) {
+            // Try as the file system resource
+            if (jsonResName.startsWith("file:")) {
+                jsonResName = jsonResName.substring(5);
             }
-            try (BufferedReader buffer = new BufferedReader(new InputStreamReader(is))) {
-                return buffer.lines().collect(Collectors.joining("\n"));
+            try {
+                is = new FileInputStream(jsonResName);
+            } catch (FileNotFoundException e) {
+                //continue
             }
+        }
+        if (is == null) {
+            throw new JwtException("Failure to open the input stream from " + jsonResName);
+        }
+        try (BufferedReader buffer = new BufferedReader(new InputStreamReader(is))) {
+            return buffer.lines().collect(Collectors.joining("\n"));
         } catch (IOException ex) {
-            throw new JwtException("Failure to read the json content:" + ex, ex);
+            throw new JwtException("Failure to read the json content from " + jsonResName + ":" + ex, ex);
         }
     }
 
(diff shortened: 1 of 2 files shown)


radcortez:
@sberyozkin Thanks!

I think we also have the same issue with `static Key getSigningKeyFromConfig(String kid)` in Quarkus native mode. What do you think?

sberyozkin:
@radcortez That issue is about the dev mode having a different class loader (key is on the classpath) while in this case it is a file system resource but I suppose loading a key from the file system should also be supported, makes sense to fix it with this PR

radcortez:
I do get this in native mode btw:

```
Caused by: java.lang.IllegalArgumentException: Signing key can not be loaded from: privateKey.pem
	at io.smallrye.jwt.build.impl.JwtSigningUtils.getSigningKeyFromConfig(JwtSigningUtils.java:346)
	at io.smallrye.jwt.build.impl.JwtSigningUtils.signJwtClaimsInternal(JwtSigningUtils.java:170)
	at io.smallrye.jwt.build.impl.JwtSignatureImpl.sign(JwtSignatureImpl.java:66)
	at com.microprofile.samples.services.book.auth.TokenUtils.generateTokenString(TokenUtils.java:31)
	at com.microprofile.samples.services.book.auth.TokenEndpoint.token(TokenEndpoint.java:31)
	at java.lang.reflect.Method.invoke(Method.java:566)
```

radcortez:
Ah, wait a second. I think I need to pass in a `resources-config.json` so the keys are included in the native image.

sberyozkin:
@radcortez OK, lets deal with it in a separate issue. I think it is https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/RSA.20Factory.20not.20available.20in.20the.20native.20mode, but may be something else. See https://github.com/quarkusio/quarkus/issues/8881 (please comment either there or may be in Zulip), we definitely don't want to ask users to prepare the graalvm configuration files manually :-).

In scope of this PR I will update the code to load the keys from the file system

sberyozkin:
@radcortez Please also pick up #232 - may be it needs to be picked up. If you enable the tracing of `KeyLocationResolver` then you will see if it is RSA factory related or not

sberyozkin:
@radcortez For example, https://github.com/quarkusio/quarkus-quickstarts/blob/master/security-jwt-quickstart/src/main/resources/application.properties, we don't add it explicitly as a graalvm resource, it works in Native (except for the recent RSA factory issue)

radcortez:
Passing the resource in GraalVM does get rid of the exception, but I do get an empty token it seems (empty String). I've also tried to pass in an http based key resource with the same result.

sberyozkin:
@radcortez I haven't concentrated on loading the keys from all sort of locations for signing the keys, in fact I probably excluded HTTP(S) out of it completely as I don't want to fetch private keys over HTTPS. Classpath is not ideal but at least one can imagine someone uses Vault etc.
Realistically a (secure) file system is the best in this case.
So as I said, in scope of this issue I'll update the code to load the keys from the filesystem. 

In meantime there is another issue in that in the native mode you don't get the private key loaded. That needs be investigated outside of this issue. As I said, it may be depending on #232, it may be the recent RSA factory issue.

radcortez:
Actually, loading the key by `http` does work, because it delegates to `static String readKeyContent(String keyLocation) throws IOException` that supports loading keys from multiple locations, including `http`. I was trying to use it as a workaround because of the resource issue in GraalVM.

They key actually seems to load properly, but you do get an empty token (apparently) and no exception. I do need to dig deeper into this.

sberyozkin:
@radcortez I forgot :-). Yeah I did to ensure it is in sync with the way the verify key can be loaded for it to be spec-ed. Sure, please have a look. Do you use a stock JDK with the GraalVM distro or as I do it [here](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/RSA.20Factory.20not.20available.20in.20the.20native.20mode/near/196791951) ?

So then as far as this PR is concerned, which is about loading a token resource from the file system, how does it look ?
Also CC @MikeEdgar

radcortez:
Yes, I use AdoptOpenJDK 11 in JAVA_HOME and GraalVM 19.3.1 for native.

Sorry, let me do a quick review.

sberyozkin:
@radcortez Thanks, will merge soon unless Mike has something to add

sberyozkin:
@MikeEdgar Thanks, yes, in time both the keys and the token resources would be loadable from anywhere. I'll need to refactor the code, for both the key and resource loading be done by the same routine

PR #236, opened 2020-05-15
Support Quarkus Dev mode with the context without the keys

sberyozkin:
We've had numerous Quarkus related feedback in that throwing the exception if no public key location is available at the CDI Producer method time affects the developer experience badly, the latest feedback is [here](https://github.com/quarkusio/quarkus/issues/5211#issuecomment-589952165).
This PR does not change anything fundamentally, right now `KeyLocationResolver` is a central place where the keys are resolved anyway, the process which fails if the location is invalid, so if location is not set or the key is not inlined, then the same `UnresolvableKeyException` is thrown.

Note I may need to revisit the whole idea of loading the key content in `KeyLocationResolver` and try to at least pre-load it for the local resources in the producer. Https resources are trickier to preload reliably as we've had the cases in Quarkus where the OIDC server is a few seconds behind. But I'll deal with it separately. There will be some best effort to ensure early that the location exists.

Here is a [Quarkus dev test](https://github.com/sberyozkin/quarkus/commit/4fc4a453037ae3ddf27e3370f9485486b49d7f57) on my branch against this PR:
- the user has a smallrye-jwt extension without any configuration
- then in the dev mode they add the properties (in the test this is achieved by removing `#` comments)
- all starts working

### DIFF
--- a/testsuite/basic/src/test/java/io/smallrye/jwt/config/JWTAuthContextInfoProviderTest.java
+++ b/testsuite/basic/src/test/java/io/smallrye/jwt/config/JWTAuthContextInfoProviderTest.java
@@ -41,13 +41,13 @@ public class JWTAuthContextInfoProviderTest {
         JWTAuthContextInfoProvider provider = JWTAuthContextInfoProvider.createWithKeyLocation("NONE", TEST_ISS);
         Optional<JWTAuthContextInfo> info = provider.getOptionalContextInfo();
         assertNotNull(info);
-        assertTrue(!info.isPresent());
+        assertTrue(info.isPresent());
     }
 
-    @Test(expected = IllegalStateException.class)
+    @Test
     public void testDefaultGetContextInfo() {
         JWTAuthContextInfoProvider provider = JWTAuthContextInfoProvider.createWithKeyLocation("NONE", TEST_ISS);
-        provider.getContextInfo();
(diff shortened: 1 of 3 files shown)


testsuite/basic/src/test/java/io/smallrye/jwt/config/JWTAuthContextInfoProviderTest.java
```diff
@@ -41,13 +41,12 @@ public void testDefaultGetOptionalContextInfo() {
         JWTAuthContextInfoProvider provider = JWTAuthContextInfoProvider.createWithKeyLocation("NONE", TEST_ISS);
         Optional<JWTAuthContextInfo> info = provider.getOptionalContextInfo();
         assertNotNull(info);
-        assertTrue(!info.isPresent());
+        assertTrue(info.isPresent());
     }
 
-    @Test(expected = IllegalStateException.class)
     public void testDefaultGetContextInfo() {
```
MikeEdgar, line 47: @sberyozkin - I think `@Test` is missed here.

sberyozkin: @MikeEdgar I wonder why it passed so easily :-)

sberyozkin:
@MikeEdgar @radcortez thanks

PR #237, opened 2020-05-15
Support for creating tokens from JsonObject

sberyozkin:
This PR supports a case where a JSON is already available, example, read from DB etc. It will make it even simpler to turn any custom JSON into custom JWT claims. 
Next I'll add more configuration properties so that 'technical'/standard claims like `issuer` and `exp` or headers like `kid` can be configured plus a set of shortcuts like `Jwt.sign(jsonObject)`


sberyozkin:
@MikeEdgar @radcortez thanks

PR #240, opened 2020-05-19
Support MP JWT 1.2 new configurations

radcortez:


sberyozkin:
Hi Roberto @radcortez This looks very neat, but in context of this issue I don't think we need to keep carrying `smallrye.jwt` properties forward, they need to be deprecated and then dropped once we switch to `mp.jwt.*` equivalents, otherwise we'll have both `smallrye.jwt.token.header` and `mp.jwt.token.header` and there is no value in it... thanks

radcortez:
We can change this to use a fallback approach. We rename the current configs and fallback to the SmallRye ones for a couple of versions and then remove them. We can even log them and warn the user that they are deprecated.

sberyozkin:
@radcortez That sounds reasonable, but I'd consider doing it in the same place where other init properties are checked, in `JWTAuthContextInfoProvider` optional producer method, using a simple `if` check: if `mp.jwt.token.header` is set - use it else 2) if `smallrye.jwt.token.header` is set - use it but warn it is deprecated and that `mp.jwt.token.header` should be used.

radcortez:
So, you don't want to use the SmallRye Config features for relocation and fallback? :)

sberyozkin:
@radcortez I'm sure they are good and can/will become handy in some contexts, but do we really need to start introducing factories, interceptors ? See, this will create several areas where we will be checking the properties, there is no convincing reason for it IMHO. It is a cool feature, no doubt, but it won't be better at this stage than continuing with the `primitive` way of checking the properties as suggested

sberyozkin:
@radcortez I think the users who deal with the custom properties migrations will love it, but in context of smallrye-jwt, may be we need a stronger reason for using the new way of doing things...

radcortez:
Ok, no worries, I'll redo this.

radcortez:
With SmallRye Config was way cleaner :D

sberyozkin:
@radcortez Thanks, this has to be done against an mpjwt12 branch. And indeed it does look more complex than it should be. Let me comment more

README.adoc
```diff
@@ -44,8 +44,10 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |Property Name|Default|Description
 |smallrye.jwt.verify.algorithm|`RS256`|Signature algorithm. Set it to `ES256` to support the Elliptic Curve signature algorithm.
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
-|smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
-|smallrye.jwt.token.cookie|none|Name of the cookie containing a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
+|mp.jwt.token.header
```
sberyozkin, line 47: There is a section somewhere above where mp jwt properties are listed, so please add them there

README.adoc
```diff
@@ -44,8 +44,10 @@ Smallrye JWT supports many properties which can be used to customize the token p
...
+(deprecated) smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
+|mp.jwt.token.cookie
+(deprecated) smallrye.jwt.token.cookie|`Bearer`|Name of the cookie containing a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
```
sberyozkin, line 50: Here lets instead have `This property is deprecated, use 'mp.jwt.... instead.'` at the end of the description part.

implementation/src/main/java/io/smallrye/jwt/config/ConfigLogging.java
```diff
@@ -30,4 +30,8 @@
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 3004, value = "mpJwtPublicKey parsed as PEM")
     void publicKeyParsedAsPem();
-}
\ No newline at end of file
+
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 3005, value = "The config %s was replaced by %s. It may be removed in a future version.")
```
sberyozkin, line 35: I propose to rephrase it as `'%s' property is deprecated and will be removed in a future version. Use '%s ' property instead'.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -131,16 +133,32 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
+    @Deprecated
+    private String smallRyeTokenHeader;
 
     /**
      * Cookie name containing a JWT token. This property is ignored unless the "smallrye.jwt.token.header" is set to 'Cookie'
      */
     @Inject
-    @ConfigProperty(name = "smallrye.jwt.token.cookie")
+    @ConfigProperty(name = "mp.jwt.token.cookie")
     private Optional<String> tokenCookie;
```
sberyozkin, line 152: Lets have `String` here, @MikeEdgar spotted we do a lot of of Opitonal with the default values, my fault really :-)

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -384,14 +412,24 @@ public String getMpJwtIssuer() {
...
-    public String getTokenHeader() {
+    public Optional<String> getTokenHeader() {
         return tokenHeader;
     }
 
     public Optional<String> getTokenCookie() {
         return tokenCookie;
     }
 
+    @Deprecated
```
sberyozkin, line 423: Lets drop these two methods, we don't really need them

testsuite/tck/pom.xml
```diff
@@ -101,6 +101,7 @@
           <skip>${skipTck}</skip>
           <suiteXmlFiles>
             <suiteXmlFile>${project.build.directory}/test-classes/tck-base-suite.xml</suiteXmlFile>
+            <suiteXmlFile>${project.build.directory}/test-classes/smallrye-suite.xml</suiteXmlFile>
```
sberyozkin, line 104: We don't need it

README.adoc
radcortez: Ok. We still need to mention that these are deprecated.

README.adoc
radcortez: Ok. Btw, we already have these in the docs, so we now have 2 locations with the configs. Can we get rid of these section in the README and point to the docs?

implementation/src/main/java/io/smallrye/jwt/config/ConfigLogging.java
radcortez: Ok.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Actually, when I was looking into this, I was thinking that we should refactor this. I don't think we have any reason to use the injection style to retrieve Configs. We should use the programmatic API. This should simplify the code greatly and allows us to test the different configurations without CDI and without any wrappers.

What do you think?

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Ok. I think we are not using the other getters as well. So maybe one more reason to refactor this :)

README.adoc
sberyozkin: Yeah, makes sense.

testsuite/tck/pom.xml
radcortez: Well, I don't feel comfortable in doing changes that should be retro compatible without any tests :)

radcortez:
> @radcortez Thanks, this has to be done against an mpjwt12 branch. And indeed it does look more complex than it should be. Let me comment more

Since this was a retro change, more merging later, but I can do it in the other branch.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: @radcortez this would take things into a completely new direction, lets discuss it separately

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: They are for Thorntail, some of them, so I'm just keep adding getters in case it ever needs them

testsuite/tck/pom.xml
sberyozkin: OK, lets just move it into testsuite/basic

sberyozkin:
@radcortez Re the original solution, this is why I said the users would love it :-), IMHO they'd appreciate it,  but in case of smallrye-jwt it just, well, complicates things IMHO, we need a simple way to control things at the library implementation level IMHO.

sberyozkin:
@radcortez I propose you update Quarkus docs with the example how to do the user-level properties migration with the latest config feature, they'd appreciate it, no doubt about it :-)

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Yes, it was not to be done in the context of this change, but something to think about or discuss. I can open a separate issue.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Ok.

testsuite/tck/pom.xml
radcortez: Ok :)

radcortez:
> @radcortez I propose you update Quarkus docs with the example how to do the user-level properties migration with the latest config feature, they'd appreciate it, no doubt about it :-)

Yeah, I have a PR opened atm for the implementation changes. When that is merged I'll work on docs :)

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: I think we need the Optionals, so we can issue the log warning messages. If we just set a simple String with the default value, we don't know if we are getting the value set by the user or the default. If we cannot use it without the default either, because it will fail in the injection part.

testsuite/tck/pom.xml
radcortez: Ok, I've had another idea and removed this. I've just implemented a test to make sure that the values match if you use different config properties. The behaviour will be tested by the TCK.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
MikeEdgar: `Optional` makes sense in the absence of a default specified in the config annotation. It's just when optional is used with a default that it becomes unnecessary.

radcortez: Yeah, for this case we do have defaults, but since the config may originate from two different properties, we cannot really use it, because then we won't know if the config was actually set or not to verify the other, unless we check the value to be the default :)

sberyozkin:
Hey Roberto, see https://github.com/smallrye/smallrye-jwt/commit/861324d91c94495213b5226ab7fc4916179fc81c#r39347882, I think we need to keep the tests which have not been yet provided for excluded to get a clean build, obviously the merge to master would have to get all the tests passed. Otherwise I've no idea if this PR is 100% correct

radcortez:
> Hey Roberto, see [861324d#r39347882](https://github.com/smallrye/smallrye-jwt/commit/861324d91c94495213b5226ab7fc4916179fc81c#r39347882), I think we need to keep the tests which have not been yet provided for excluded to get a clean build, obviously the merge to master would have to get all the tests passed. Otherwise I've no idea if this PR is 100% correct

Correct. Let me do it.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: @radcortez IMHO we don't need `Optional`, I don't really understand what it is to do with the config originating from the 2 different properties.

doc/modules/ROOT/pages/configuration.adoc
```diff
@@ -33,8 +36,8 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |Property Name|Default|Description
 |smallrye.jwt.verify.algorithm|`RS256`|Signature algorithm. Set it to `ES256` to support the Elliptic Curve signature algorithm.
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
-|smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
-|smallrye.jwt.token.cookie|none|Name of the cookie containing a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
+|smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.This property is deprecated. Use `mp.jwt.token.header` instead.
```
sberyozkin, line 39: Minor one, please add a space before `This`

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -160,16 +164,41 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
     @ConfigProperty(name = "mp.jwt.verify.requireiss", defaultValue = "true")
     private Optional<Boolean> mpJwtRequireIss;
 
+    /**
+     * @since 1.2
+     */
+    @Inject
+    @ConfigProperty(name = Names.TOKEN_HEADER)
+    private Optional<String> mpJwtTokenHeader;
```
sberyozkin, line 172: Both this and then next property should be just Strings with the default properties

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -160,16 +164,41 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
     // SmallRye JWT specific properties
     /**
      * HTTP header which is expected to contain a JWT token, default value is 'Authorization'
+     *
+     * @deprecated Use {@link JWTAuthContextInfoProvider#mpJwtTokenHeader}
      */
     @Inject
-    @ConfigProperty(name = "smallrye.jwt.token.header", defaultValue = AUTHORIZATION_HEADER)
-    private String tokenHeader;
+    @ConfigProperty(name = "smallrye.jwt.token.header")
```
sberyozkin, line 195: Why are removing the defaults ?

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: @radcortez I got it :-), let me remove the other comments

sberyozkin: Yeah, we will get rid op Optional once we drop eventually the newly deprecated properties

doc/modules/ROOT/pages/configuration.adoc
sberyozkin: Never mind, I will do it later

sberyozkin:
@MikeEdgar Hi Mike, so as discussed, we'll drop Optional later once we drop the newly deprecated properties

PR #241, opened 2020-05-19
Throw NotAuthorizedException if a JWT is sent to an unauthenticated endpoint and is not valid

radcortez:
To fix TCK test EmptyTokenTest in MP 1.2.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/jaxrs/JWTAuthenticationFilter.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/jaxrs/JWTAuthenticationFilter.java
@@ -21,11 +21,13 @@ import java.security.Principal;
 
 import javax.annotation.Priority;
 import javax.inject.Inject;
+import javax.ws.rs.NotAuthorizedException;
 import javax.ws.rs.Priorities;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.container.PreMatching;
 import javax.ws.rs.core.Cookie;
+import javax.ws.rs.core.Response;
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/jaxrs/JWTAuthenticationFilter.java
```diff
@@ -62,6 +61,7 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
 
             if (bearerToken != null) {
                 try {
+                    requestContext.setProperty(HAS_JWT, true);
```
sberyozkin, line 64: I've looked at this filter, it catches the exception, logs, and continues, it needs to return 401 immediately, so a new property, and a new optional filter won't be needed

radcortez: The idea was to follow the same structure we had with the others filters. Sure, I can change it.

sberyozkin: Yeah, if we have a token and it failed to parse then relying on some follow up filters to block such an invalid token will not be very safe...

sberyozkin: @radcortez In quarkus we may need to relax things a bit, here we have a proactive authentication, but in Quarkus Stuart added an option to authenticate lazily, but we can discuss it later in Quarkus

radcortez: Sure. Quarkus doesn't use this filter anyway.

PR #245, opened 2020-05-22
Fixed issues with sign and decrypt.

radcortez:


implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -166,6 +166,10 @@ private String decryptSignedToken(String token, JWTAuthContextInfo authContextIn
                             KeyEncryptionAlgorithm.RSA_OAEP.getAlgorithm()));
             jwe.setKey(getDecryptionKeyResolver(authContextInfo).resolveKey(jwe, null));
             jwe.setCompactSerialization(token);
+            if (!"JWT".equals(jwe.getContentTypeHeaderValue())) {
```
sberyozkin, line 169: @radcortez We don't need it, the mp text currently requires a JWT header somewhere but this headers does not bring any value in reality

sberyozkin: @radcortez Oh sorry this is the internal token

sberyozkin: No, we don't need still, on the builder API side we set this header but here it is too strict to enforce it

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -363,7 +363,9 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
             contextInfo.setPublicKeyContent(mpJwtPublicKey.get());
         } else if (mpJwtLocation.isPresent() && !NONE.equals(mpJwtLocation.get())) {
             contextInfo.setPublicKeyLocation(mpJwtLocation.get().trim());
-        } else if (mpJwtDecryptKeyLocation.isPresent() && !NONE.equals(mpJwtDecryptKeyLocation.get())) {
```
sberyozkin, line 366: Was it a rebase issue ? Probably, thanks for fixing it

radcortez: Not sure what it was, but it was failing one of the tests :)

implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
radcortez: This is to verify if the Content Type header is part of the token. If not, we fail the TCK test `org.eclipse.microprofile.jwt.tck.container.jaxrs.jwe.RolesAllowedSignEncryptTest#callEchoWithoutCty. 

Since we unwrap the Token before passing it to JOSE, it doesn't perform the content validation.

sberyozkin: @radcortez I forgot :-), OK, lets start with enforcing the best practice

PR #246, opened 2020-05-22
Minor JWT build improvements

sberyozkin:
This PR:
* Deprecates a number of builder API methods and introduces more compact alternatives.

  For example, when we do:
  `Jwt.claims("somejson.json").jws().signatureKeyId("kid").sign()`, `signature` in `signatureKeyId` is duplicating what both `jws()` and `sign` imply. Similarly for all other deprecated methods. The IDE also helps with algorithm related methods, for ex, when we do `jws().algorithm(...)` we see a `SignatureAlgorithm` parameter type.

  In fact, originally I did plan for `Jwt.claims("somejson.json").jws().keyId("kid").sign()` - the compactness of the API is a big priority, but then I somehow got offtrack when I was looking at the inner token case, where the token is signed and encrypted:
```
Jwt.claims().jws().keyId().innerSign().keyId().encrypt();
```
Here I thought well, 2 `keyId()`, 1st one sets a key if of the signing key, 2nd one - of the key encryption key.
But over time I've come to the conclusion that this case is quite advanced, the mainstream case is `sign` and in this advanced case, after the formatting, it also becomes more readable. `kid`s will likely be configurable too.
* deprecated `none` for the `innerSign()` - I came to this conclusion while doing the initial MP JWT PR, it is basically useless :-) to have encrypt() with an internal `none` as it does not add anything to the `encrypt` only case, i.e, it does not help with identifying who signed it. It also makes it hard to spot if the user made a typo in the config property

All of these deprecated methods/properties will be dropped in 3.0.0

* config property support for `exp` and `issuer` - these are the only 2 claims that are required by MP JWT, so it will make much easier to test, every time I see a custom `exp` being set in the code I'm thinking, oh, we need to get it configured.

### DIFF
--- a/doc/modules/ROOT/pages/generate-jwt.adoc
+++ b/doc/modules/ROOT/pages/generate-jwt.adoc
@@ -34,11 +34,21 @@ JwtClaimsBuilder builder2 = Jwt.claims("/tokenClaims.json");
 
 // Builder created from a map of claims
 JwtClaimsBuilder builder3 = Jwt.claims(Collections.singletonMap("customClaim", "custom-value"));
+
+// Builder created from JsonObject
+JsonObject userName = Json.createObjectBuilder().add("username", "Alice").build();
+JsonObject userAddress = Json.createObjectBuilder().add("city", "someCity").add("street", "someStreet").build();
+JsonObject json = Json.createObjectBuilder(userName).add("address", userAddress).build();
+JwtClaimsBuilder builder4 = Jwt.claims(json);
+
+// Builder created from JsonWebToken
+@Inject JsonWebToken token;
+JwtClaimsBuilder builder5 = Jwt.claims(token);
 ----
 
 The API is fluent so the builder initialization can be done as part of the fluent API sequence. The builder will also
-set `iat (issued at) to the current time, `exp`(expires at) to 5 minutes away from the current time and `jti`
-(unique token identifier) claims if they have not already been set, so one can skip setting them when possible.
+set `iat (issued at) claim to the current time, `exp`(expires at) claim to a sum of the `iat` claim and `smallrye.jwt.new-token.lifespan`
+property values and `jti` (unique token identifier) claim if they have not already been set, so one can skip setting them when possible.
 
 The next step is to decide how to secure the claims.
 
@@ -56,7 +66,7 @@ import io.smallrye.jwt.build.Jwt;
 String jwt1 = Jwt.claims("/tokenClaims.json").sign();
 
 // Set the headers and sign the claims with an RSA private key loaded in the code (the implementation of this method is omitted). Note a 'jws()' transition to a 'JwtSignatureBuilder'.
-String jwt2 = Jwt.claims("/tokenClaims.json").jws().signatureKeyId("kid1").header("custom-header", "custom-value").sign(getPrivateKey());
+String jwt2 = Jwt.claims("/tokenClaims.json").jws().keyId("kid1").header("custom-header", "custom-value").sign(getPrivateKey());
 ----
 
 Note the `alg` (algorithm) header is set to `RS256` by default.
@@ -102,6 +112,8 @@ Smallrye JWT supports the following properties which can be used to customize th
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
-|smallrye.jwt.encrypt.key-location|none|Config property allows the location of a key to be specified which will be used to encrypt the claims or inner JWT.
-|smallrye.jwt.sign.key-location|none|Config property allows the location of a private key to be specified which will be used to sign the claims of a JWT.
+|smallrye.jwt.encrypt.key-location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called.
+|smallrye.jwt.sign.key-location|none|Location of a key which will be used to sign the claims when either a no-argument sign() or innerSign() method is called.
+|smallrye.jwt.new-token.lifespan|300|Token lifespan in seconds which will be used to calculate an `exp` (expiry) claim value if this claim has not already been set.
+|smallrye.jwt.new-token.issuer|none|Token issuer which can be used to set an `iss` (issuer) claim value if this claim has not already been set.
 |===
--- a/implementation/src/main/java/io/smallrye/jwt/build/JwtEncryptionBuilder.java
+++ b/implementation/src/main/java/io/smallrye/jwt/build/JwtEncryptionBuilder.java
@@ -22,27 +22,75 @@ public interface JwtEncryptionBuilder extends JwtEncryption {
      * Note that only 'RSA-OAEP-256' (default), 'ECDH-ES+A256KW' and 'A256KW' algorithms must be supported.
      * A key of size 2048 bits or larger MUST be used with 'RSA-OAEP-256' algorithm.
      * 
+     * @since 2.1.3
+     *
      * @param algorithm the key encryption algorithm
      * @return JwtEncryptionBuilder
      */
-    JwtEncryptionBuilder keyEncryptionAlgorithm(KeyEncryptionAlgorithm algorithm);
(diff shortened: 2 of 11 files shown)


implementation/src/main/java/io/smallrye/jwt/build/JwtEncryptionBuilder.java
```diff
@@ -25,7 +25,29 @@
...
+     * Note that only 'RSA-OAEP-256' (default), 'ECDH-ES+A256KW' and 'A256KW' algorithms must be supported.
+     * A key of size 2048 bits or larger MUST be used with 'RSA-OAEP-256' algorithm.
+     *
+     * @param algorithm the key encryption algorithm
+     * @return JwtEncryptionBuilder
+     */
+    @Deprecated
+    default JwtEncryptionBuilder keyEncryptionAlgorithm(KeyEncryptionAlgorithm algorithm) {
+        return keyAlgorithm(algorithm);
+    }
```
MikeEdgar, line 41: Maybe add `@deprecated` to the JavaDoc also with a reference to the replacement method. Same comment for the other deprecated methods.

implementation/src/main/java/io/smallrye/jwt/build/JwtEncryptionBuilder.java
```diff
@@ -25,7 +25,29 @@
...
+
+    /**
+     * Set an 'alg' key encryption algorithm.
+     * Note that only 'RSA-OAEP-256' (default), 'ECDH-ES+A256KW' and 'A256KW' algorithms must be supported.
+     * A key of size 2048 bits or larger MUST be used with 'RSA-OAEP-256' algorithm.
+     *
+     * @param algorithm the key encryption algorithm
+     * @return JwtEncryptionBuilder
+     */
+    @Deprecated
```
radcortez, line 38: Can we add the tag to point to the method that should be used? It should be straightforward, but it would hurt :)

implementation/src/main/java/io/smallrye/jwt/build/JwtEncryptionBuilder.java
```diff
@@ -34,15 +56,29 @@
      * @param algorithm the content encryption algorithm
      * @return JwtEncryptionBuilder
      */
-    JwtEncryptionBuilder contentEncryptionAlgorithm(ContentEncryptionAlgorithm algorithm);
+    @Deprecated
```
radcortez, line 59: Same as above.

implementation/src/main/java/io/smallrye/jwt/build/JwtEncryptionBuilder.java
```diff
@@ -34,15 +56,29 @@
...
+    JwtEncryptionBuilder keyId(String keyId);
 
     /**
      * Set a 'kid' key encryption key id.
      * 
      * @param keyId the key id
      * @return JwtEncryptionBuilder
      */
-    JwtEncryptionBuilder keyEncryptionKeyId(String keyId);
+    @Deprecated
```
radcortez, line 78: Same as above.

implementation/src/main/java/io/smallrye/jwt/build/JwtSignatureBuilder.java
```diff
@@ -19,22 +19,45 @@
...
+
     /**
      * Set a signature algorithm.
      * Note that only 'RS256' (default), 'ES256' and 'HS256' algorithms must be supported.
      * 
      * @param algorithm the signature algorithm
      * @return JwtSignatureBuilder
      */
-    JwtSignatureBuilder signatureAlgorithm(SignatureAlgorithm algorithm);
+    @Deprecated
```
radcortez, line 38: Same as above

implementation/src/main/java/io/smallrye/jwt/build/JwtSignatureBuilder.java
```diff
@@ -19,22 +19,45 @@
...
+    JwtSignatureBuilder keyId(String keyId);
 
     /**
      * Set a 'kid' signature key id
      * 
      * @param keyId the key id
      * @return JwtSignatureBuilder
      */
-    JwtSignatureBuilder signatureKeyId(String keyId);
+    @Deprecated
```
radcortez, line 57: Same as above.

implementation/src/main/java/io/smallrye/jwt/build/impl/JwtSigningUtils.java
```diff
@@ -208,12 +208,24 @@ static void setDefaultJwtClaims(JwtClaims claims) {
...
+        }
+    }
+
+    static <T> T getConfigProperty(String name, Class<T> cls) {
+        try {
+            return ConfigProvider.getConfig().getValue(name, cls);
+        } catch (NoSuchElementException ex) {
+            return null;
+        }
     }
```
MikeEdgar, line 229: It might be cleaner to have this method take a default value, then use `getOptionalValue(name, cls).orElse(defaultValue)`. You can avoid the `try`/`catch` and also the conditional logic for setting the timeout of `300L`

implementation/src/test/java/io/smallrye/jwt/build/JwtBuildConfigSource.java
```diff
@@ -8,6 +8,8 @@
 public class JwtBuildConfigSource implements ConfigSource {
 
     boolean signingKeyAvailable = true;
+    boolean lifespanPropertyRequered;
+    boolean issuerPropertyRequered;
```
MikeEdgar, line 12: Just a test class, but required is spelled incorrectly ;-)

sberyozkin:
Hi @MikeEdgar @radcortez thanks, will deal with the comments

implementation/src/main/java/io/smallrye/jwt/build/impl/JwtSigningUtils.java
sberyozkin: Hi Mike @MikeEdgar I've tried to optimize a bit, as indeed it is only in the case of the timeout that a non null default value is available. Please check

sberyozkin:
@MikeEdgar @radcortez I think I've addressed all the comments, cheers

sberyozkin:
Hi Mike thanks, I've realized I forgot to update the docs (also updated the text related to the previous changes in sync with the similar Quarkus text, @bobbyphilip - minor update around the key properties to clarify when they are effective) - please check, I can merge later this evening if there will be no comments but if some wording/grammar can be improved then it will be fine.  
CC @radcortez

doc/modules/ROOT/pages/generate-jwt.adoc
```diff
@@ -102,6 +112,8 @@ Smallrye JWT supports the following properties which can be used to customize th
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
-|smallrye.jwt.encrypt.key-location|none|Config property allows the location of a key to be specified which will be used to encrypt the claims or inner JWT.
-|smallrye.jwt.sign.key-location|none|Config property allows the location of a private key to be specified which will be used to sign the claims of a JWT.
+|smallrye.jwt.encrypt.key-location|none|Location of a public key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called.
```
bobbyphilip, line 115: To be pedantic, this can be either a public key or a secret key. The encrypt()  API supports both

sberyozkin: @bobbyphilip Yeah, I forgot it will work if JWK format is used, I'll just update to a `signing key`, `encrypting key` for both properties...

sberyozkin:
@radcortez Thanks, I'll do another couple of minor updates :-)

doc/modules/ROOT/pages/generate-jwt.adoc
sberyozkin: @bobbyphilip Or simply `a key` as per your original text

sberyozkin:
Merging now as it has already been approved a few times, updated the text to refer to a `key` only as opposed to a `public key` or `private key`.

PR #247, opened 2020-05-22
Support mp.jwt.verify.publickey.algorithm.

radcortez:
This requires https://github.com/eclipse/microprofile-jwt-auth/pull/190.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
+++ b/implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
@@ -26,7 +26,6 @@ import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 
 import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.eclipse.microprofile.jwt.config.Names;
 import org.jboss.logging.Logger;
 
 import io.smallrye.jwt.KeyFormat;
@@ -98,6 +97,7 @@ public class JWTAuthContextInfoProvider {
         JWTAuthContextInfoProvider provider = new JWTAuthContextInfoProvider();
         provider.mpJwtPublicKey = Optional.of(publicKey);
         provider.mpJwtLocation = Optional.of(publicKeyLocation);
+        provider.mpJwtPublicKeyAlgorithm = Optional.of(SignatureAlgorithm.RS256);
         provider.mpJwtDecryptKeyLocation = Optional.of(decryptKeyLocation);
         provider.mpJwtIssuer = issuer;
         provider.mpJwtRequireIss = Optional.of(Boolean.TRUE);
@@ -119,9 +119,9 @@ public class JWTAuthContextInfoProvider {
         provider.maxTimeToLiveSecs = Optional.empty();
         provider.jwksRefreshInterval = Optional.empty();
         provider.forcedJwksRefreshInterval = 30;
-        provider.signatureAlgorithm = Optional.of(SignatureAlgorithm.RS256);
+        provider.signatureAlgorithm = provider.mpJwtPublicKeyAlgorithm;
         provider.keyFormat = KeyFormat.ANY;
-        provider.expectedAudience = Optional.empty();
+        provider.expectedAudience = provider.mpJwtVerifyAudiences;
         provider.groupsSeparator = DEFAULT_GROUPS_SEPARATOR;
         provider.requiredClaims = Optional.empty();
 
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -143,6 +144,13 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
     @ConfigProperty(name = "mp.jwt.verify.publickey.location", defaultValue = NONE)
     private Optional<String> mpJwtLocation;
 
+    /**
+     * @since 1.2
```
sberyozkin, line 148: @radcortez  I believe `1.1`, `1.2` etc should only be in the MP JWT source, here it should be `3.0.0` as not all the new methods/properties we will be adding will be MP JWT rellated so we should use smallrye-jwt versions in `since` tags ?

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -430,16 +444,26 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
         contextInfo.setMaxTimeToLiveSecs(maxTimeToLiveSecs.orElse(null));
         contextInfo.setJwksRefreshInterval(jwksRefreshInterval.orElse(null));
         contextInfo.setForcedJwksRefreshInterval(forcedJwksRefreshInterval);
-        if (signatureAlgorithm.orElse(null) == SignatureAlgorithm.HS256) {
-            throw ConfigMessages.msg.hs256NotSupported();
+        if (mpJwtVerifierPublicKeyAlgorithm.isPresent()) {
```
sberyozkin, line 447: I propose to optimize this block a bit, have an in scope property set to either mp jwt or smallrye prop, and do HS256 check once

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -143,6 +144,13 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
     @ConfigProperty(name = "mp.jwt.verify.publickey.location", defaultValue = NONE)
     private Optional<String> mpJwtLocation;
 
+    /**
+     * @since 1.2
+     */
+    @Inject
+    @ConfigProperty(name = Names.VERIFIER_PUBLIC_KEY_ALGORITHM)
+    private Optional<SignatureAlgorithm> mpJwtVerifierPublicKeyAlgorithm;
```
sberyozkin, line 152: I'm not sure using Names helps actually, I'd like to see the name of the property by looking at the source here, besides Names `Verifier` does not match `verify` in `mp.jwt.verify`, I probably used `Names.Verifier` because this pattern was already there. Anyway, may be just lets have `mpJwtPublicKeyAlgorithm` similarly to `mpJwtPubicKeyLocation` to keep it shorter a bit

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: I thought we were using the MP API version for properties coming from the spec. At least others properties only reference 1.0 and 1.1. So we should probably make it consistent across the board.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Ok

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
radcortez: Ok, I can change it to use the plain String. I thought it was better to use the static name so it would be easier for search.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: @radcortez Yeah, it was originally there, but I guess it was not correct. For example, we have several `since 2.1.3` tags now so having `1.0`, `2.1.3`. I propose drop all `1.0`, `1.1`, etc and only refer to smallrye versions, and may be add some plain Java doc comments that a given property is part of MP JWT 1.2, etc.

implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
sberyozkin: Please, it feels it will be easier to read the source. And please have either `mpJwtPublicKeyAlgorithm` or `mpJwtVerifyPublicKeyAlgorithm`, I'd go for the former as we currently don't use `verify` in other similar properties and don't have to match the internal property names one to one with the spec property names...

sberyozkin:
@radcortez `apiNote`, that works :-)

PR #249, opened 2020-05-25
Refactor JWTEncryptionImpl to remove redundant code paths

bobbyphilip:
A slight refactor of the JWTEncryptionImpl.
 Merged some of the code paths
 Corrected the Exception named in the ImplMessage to match that of the API

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/build/JwtEncryption.java
+++ b/implementation/src/main/java/io/smallrye/jwt/build/JwtEncryption.java
@@ -53,8 +53,8 @@ public interface JwtEncryption {
      * unless different ones have been set with {@code JwtEncryptionBuilder}.
      * A key of size 2048 bits or larger MUST be used with the 'RSA-OAEP' and 'RSA-OAEP-256' algorithms.
      *
-     * @return signed JWT token
-     * @throws JwtEncryptionException the exception if the signing operation has failed
+     * @return encrypted JWT token
+     * @throws JwtEncryptionException the exception if the encryption operation has failed
      */
     String encrypt() throws JwtEncryptionException;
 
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/jwt/build/impl/JwtEncryptionImpl.java
```diff
@@ -64,15 +64,15 @@ public String encrypt(String keyLocation) throws JwtEncryptionException {
...
         }
-        return key instanceof PublicKey ? encryptInternal((PublicKey) key) : encryptInternal((SecretKey) key);
+        return encryptInternal(key);
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
     public String encrypt() throws JwtSignatureException {
```
bobbyphilip, line 74: Ideally this should be a JWTEncryptionException, but that means the interface will have to change. I am not sure if this is a breaking change

sberyozkin: @bobbyphilip It is just a bug, thanks for spotting it, please update the interface, it is still a work in progress, we are tuning and optimizing things, some refactorings and other ifxes can go freely when needed

implementation/src/main/java/io/smallrye/jwt/build/JwtEncryption.java
```diff
@@ -54,8 +54,8 @@
      * A key of size 2048 bits or larger MUST be used with the 'RSA-OAEP' and 'RSA-OAEP-256' algorithms.
      *
      * @return signed JWT token
```
sberyozkin, line 56: Hi @bobbyphilip Can you also please replace `signed' with `encrypted` and `signing` (operation) with `encryption`

PR #255, opened 2020-06-11
Improve the resource loading code

sberyozkin:
This PR does the following:
 - Moves the resource loading code into `ResourceUtils` to make it consistent across all of smallrye-jwt, so that the keys (for signing, encrypting, verifying, decrypting), as well as the claim resources (builder API) can be loaded from everywhere.
 - Started with enforcing early that `mp.jwt.publickey.location` points to the actual key(s), this in fact will be closer to the spec expectations. (I'm still keeping initializing `http(s)` locations in `KeyLocationResolver` as Keycloak/etc can be a few secs behind at the start up time, but I'll probably, in the next phase, will try to init them early as well, and if it fails - then let `KeyLocationResolver` handle it...). CC @FroMage, Stephane has asked for it after spending a few hours chasing the problem with an extra space in the location property :-). This will also make things a bit faster....

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
+++ b/implementation/src/main/java/io/smallrye/jwt/KeyUtils.java
@@ -16,16 +16,13 @@
  */
 package io.smallrye.jwt;
 
-import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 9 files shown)


sberyozkin:
@radcortez thanks

PR #256, opened 2020-06-11
Issue#253 Use a Galleon provisioned WildFly Elytron based configuration

darranl:
I suspect this may need some discussion to see if some aspects should more to smallrye parent but in general this update instead of unzipping a servlet feature pack distribution uses Galleon to dynamically provision a server with the minimal set of layers needed to run the TCK in this form.

This change also switches to using WildFly Elytron instead of the legacy security provided by PicketBox, at some point in the future we will be moving to remove PicketBox from the configurations - additionally it is know issues will be experienced with PicketBox from Java 14 and onwards.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,7 @@
     <version.weld-junit4>2.0.1.Final</version.weld-junit4>
     <version.galleon>4.2.5.Final</version.galleon>
     <version.wildfly>20.0.0.Final</version.wildfly>
+    <version.wildfly.plugin>2.0.2.Final</version.wildfly.plugin>
     <sonar.coverage.jacoco.xmlReportPaths>${project.basedir}/testsuite/basic/target/site/jacoco-aggregate/jacoco.xml,${project.basedir}/../testsuite/basic/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
   </properties>
 
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,7 @@
     <version.resteasy>3.12.1.Final</version.resteasy>
     <version.mokito>3.3.3</version.mokito>
     <version.weld-junit4>2.0.1.Final</version.weld-junit4>
+    <version.wildfly>20.0.0.Final</version.wildfly>
     <sonar.coverage.jacoco.xmlReportPaths>${project.basedir}/testsuite/basic/target/site/jacoco-aggregate/jacoco.xml,${project.basedir}/../testsuite/basic/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
   </properties>
 
@@ -129,6 +130,7 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
+
 
   <build>
     <plugins>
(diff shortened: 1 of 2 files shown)


darranl:
Looks like an issue to debug there ;-(

radcortez:
```
[ERROR] Failed to execute goal org.jboss.galleon:galleon-maven-plugin:4.2.5.Final:provision (default) on project smallrye-jwt-tck: Provisioning failed: Failed to process JBoss module XML template for feature-pack wildfly-core@maven(org.jboss.universe:community-universe):current#12.0.1.Final package org.apache.xerces: Failed to resolve artifact xerces:xercesImpl:jar:2.12.0.SP02: Failed to resolve xerces:xercesImpl:jar:2.12.0.SP02: Could not find artifact xerces:xercesImpl:jar:2.12.0.SP02 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
```

Weird, should Galleon use a SP dependency?

darranl:
I am just looking into the dependencies, I think it is actually the WildFly Core feature pack with that dependency.  An SP dependency is fine as that is just a service pack but I suspect there is an issue that most artefacts other than xerces are synchronising to maven central.

radcortez:
My doubt was mostly if SP jar's are published in Central, and if we could use them without adding extra repos?

darranl:
Yeah SPs do go to Maven Central here is an example of one of ours https://search.maven.org/artifact/org.jboss.msc/jboss-msc/1.1.7.SP1/jar

I am just wondering where the SP tag of xerces was created.

radcortez:
Ok great :) So we just need to figure out where it is :)

darranl:
I suspect this latest change will not be acceptable but am just submitting anyway as will be interesting to see if this gives a clean CI run.

Overall there are two problems:

1. We have tagged the SP release of xerces ourselves so although Maven central takes service pack releases this is not our project so is only available in our repository.
2. WildFly presently depends on a version of PicketBox in a repository hosted by Red Hat so that also does not synchronise to Maven central.

radcortez:
It does work. I'm just not sure if we should depend on these repos. @sberyozkin any thoughts?

darranl:
I am going to discuss with the WildFly team as well and see if there is anything else we could do to eliminate needing these repos.

darranl:
@sberyozkin / @radcortez I think before I rebase this one again it may be worth checking if this could gain approval.

Realistically at the moment I don't think provisioning WildFly can completely eliminate using repository.jboss.org for the provisioning.

Presently SmallRye JWT is testing against WildFly 18.0.1.Final but we have just released 23.0.0.Final, if we switch to Galleon provisioning it will hopefully be a lot easier to keep that up to date using a slimmed configuration to meet the test requirements.

This probably does also raise the question of should the WildFly version really be managed in a parent pom as I suspect individual projects may have their own issues to work through for the upgrades.

radcortez:
I'm ok with this. We probably need to remove the `wildfly.version` from the parent pom, since it may cause confusions and errors with the supported Wildfly version for each of SR implementations.

It should be fine to just set the version in the specific TCK project. If additional repositories are required, we can maybe put them in the TCK module directly. The TCK module is not being deployed to Nexus, so I think it should be fine, but we should verify this of course.

darranl:
Thanks @radcortez I will rebase and check where the repository is defined, it is for testing only so we don't need it to leak into the published artefacts.

darranl:
One further update, to rebase this branch I have also been looking to move to WildFly 23 so we are using the latest version available.  I ran into a couple of issues within Weld so we are currently discussing some fixes, once I have access to a Weld release with the fixes I need I can complete these updates.

darranl:
Closing so I can resubmit

PR #257, opened 2020-06-11
Support public RSA keys with 1024 bit length

sberyozkin:
Darran @darranl has identified the specification problem where RSA keys with the `1024` bit length are required to be supported but neither TCK enforces it nor smallrye-jwt itself supports such keys. It concerns the WildFly security team. 
Darran has proposed to drop `1024` key length completely in `MP JWT 1.2` (which is very sound since MP JWT builds on top of the Jose specs and the JWA spec in particular requiring a minimum `2048` RSA key length) but due to the pushback we will have it in MP JWT 1.2.
This PR makes sure it is supported OOB with an option for the users to block `1024` bit length keys

### DIFF
--- a/doc/modules/ROOT/pages/configuration.adoc
+++ b/doc/modules/ROOT/pages/configuration.adoc
@@ -33,6 +33,7 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |Property Name|Default|Description
 |smallrye.jwt.verify.algorithm|`RS256`|Signature algorithm. Set it to `ES256` to support the Elliptic Curve signature algorithm.
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
+|smallrye.jwt.verify.relax-key-validation|false|Relax the validattion of the verification keys, setting this property to `true` will allow public RSA keys with the length less than 2048 bit.
 |smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
 |smallrye.jwt.token.cookie|none|Name of the cookie containing a token. This property will be effective only if  `smallrye.jwt.token.header` is set to `Cookie`.
 |smallrye.jwt.always-check-authorization|false|Set this property to `true` for `Authorization` header be checked even if the `smallrye.jwt.token.header` is set to `Cookie` but no cookie with a `smallrye.jwt.token.cookie` name exists.
--- a/testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParserTest.java
+++ b/testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParserTest.java
@@ -3,16 +3,22 @@ package io.smallrye.jwt.auth.principal;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 
+import java.security.KeyPair;
 import java.security.interfaces.RSAPublicKey;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 
 import org.eclipse.microprofile.jwt.tck.util.TokenUtils;
+import org.jose4j.jwt.consumer.InvalidJwtException;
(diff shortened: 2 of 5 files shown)


radcortez:
Hum, not sure how the `relax` naming. It seems we are opening up to accept more stuff, but we actually reducing the options :)

sberyozkin:
@radcortez Hi, this is a 1 to 1 mapping with the Jose4j property, it can be tricky to name it correctly given that we may also enforce something around EC keys going forward, so something like `acceptRsaKeysWith1024Length` or similar would not be cool :-)

`relaxKeyValidation=true` is vague enough and can be doc-ed (right now) that by default we accept RSA keys < 2048 bit length.

However I've been wondering if we should continue with enforcing `2048` by default and only set this property to `true` if we have a TCK test checking `1024` bit length (in the mp jwt 1.2 branch).  

@darranl Would that work for WildFly ? No one has reported an issue around `1024` not being supported OOB...
Thanks

sberyozkin:
Actually, to avoid surprising the 2.1.x users, I think `1024` should be disabled OOB. If someone ever comes across a situation where the token signed with `1024` keys have to be accepted then they'd just enable this property. 
May be we will enable by default at the MP JWT 1.2 branch...
I'm going to reset

doc/modules/ROOT/pages/configuration.adoc
```diff
@@ -33,6 +33,7 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |Property Name|Default|Description
 |smallrye.jwt.verify.algorithm|`RS256`|Signature algorithm. Set it to `ES256` to support the Elliptic Curve signature algorithm.
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
+|smallrye.jwt.verify.relax-key-validation|true|Relax the validattion of the verification keys, setting this property to `false` will disallow public RSA keys with the length less than 2048 bit.
 |smallrye.jwt.token.header|`Authorization`|Set this property if another header such as `Cookie` is used to pass the token.
```
MikeEdgar, line 37: This might need some rewording with the change of the default from `true` to `false`.

testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParserTest.java
```diff
@@ -35,6 +40,28 @@ public void testParse() throws Exception {
...
+    public void testParseJwtSignedWith1024RsaKeyLengthDisallowed() throws Throwable {
+        KeyPair pair = KeyUtils.generateKeyPair(1024);
+        String jwt = TokenUtils.generateTokenString(pair.getPrivate(), "kid", "/Token1.json", null, null);
+        JWTAuthContextInfo context = new JWTAuthContextInfo((RSAPublicKey) pair.getPublic(), "https://server.example.com");
+        try {
+            parser.parse(jwt, context);
+            fail("InvalidJwtException is expected");
+        } catch (ParseException ex) {
+            throw ex.getCause();
+        }
```
MikeEdgar, line 62: Consider using `assertThrows` instead of `@Test(expected = ...)`. It's the more JUnit 5 way of doing things, which is probably coming in the future.

```java
ParseException thrown = assertThrows(ParseException.class, () -> parser.parse(jwt, context));
assertTrue(thrown.getCause() instanceof InvalidJwtException);
```

darranl:
@sberyozkin I think disabled by default is fine, provided we can enable to get a TCK pass.

testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParserTest.java
radcortez: Parent already has JUnit 5 support, so we could move our tests there.

sberyozkin:
Hi @darranl For 3.0.0 (mp jwt 1.2 branch) we can flip it to true if needed or may be it can be done for the TCK run only (both here and in the specific integrations), we will make it right :-).

testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParserTest.java
sberyozkin: @MikeEdgar @radcortez `assertThrows` is good, this is definitely cleaner/more compact, done

sberyozkin:
OK, thanks all, merging now

PR #258, opened 2020-06-12
JWT Builder shortcuts

sberyozkin:
This PR continues with making the builder API as compact as possible, introduces the shortcuts to optimize some flows.

1. In the flow where the claims are built from the empty builder, `Jwt.claims()` transition is now optional, every time I see `Jwt.claims().claim("a", "b")` I think this is suboptimal. 
So, instead of `Jwt.claims().issuer("issuer").subject("sub").sign()` once can just do `Jwt.issuer("issuer").subject("sub").sign()`.

2. Creates `sign()`, `encrypt` and `innerSignAndEncrypt` shortcuts to optimize the flows where the keys are configured and where the default RSA algorithms are used:
 - `Jwt.claims("/token.json").sign()` is equivalent to `Jwt.sign("/token.json")`
(same for the maps and jsonobject)
- `Jwt.claims("/token.json").jwe().encrypt()` is equivalent to `Jwt.encrypt("/token.json")`
(same for the maps and jsonobject)
- `Jwt.claims("/token.json").innerSign().encrypt()` is equivalent to `Jwt.innerSignAndEncrypt("/token.json")`
(same for the maps and jsonobject)

3. Also added `Instant` overloads for the `iss` and `exp` claims

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/build/Jwt.java
+++ b/implementation/src/main/java/io/smallrye/jwt/build/Jwt.java
@@ -2,6 +2,7 @@ package io.smallrye.jwt.build;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Set;
 
 import javax.json.JsonArray;
 import javax.json.JsonObject;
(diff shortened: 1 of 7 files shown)


sberyozkin:
@MikeEdgar @radcortez thanks

PR #263, opened 2020-06-17
Deprecate JwtClaimsBuilder.json, rename JwtSigningUtils to JwtBuilderUtils, create JwtProvider once

sberyozkin:
Hi All, This PR is another round of the builder API cleanup:

- `JwtClaimsBuilder.json()` is deprecated. It can be useful to log what is about to be signed, but `json()` has side-effects (adding the default claims if any) and I'm doubting now it can be done when someone just wants to log, the default claims if any should only be added with the flow terminating methods like `sign`/etc. I can probably add a property to require that the claims are logged just before they are about to be signed.
- Renamed `JwtSigningUtils` to `JwtBuildUtils`. Some history: I started the JWT generation work with prototyping `JwtSigningUtils` and then decided to move to creating a proper API, `JwtSigningUtils` got moved to `build.impl` subpackage to stress it was really an internal helper class. Some methods like `setDefaultClaims` is called from the encryption code, and it ships a lot of untested and undocumented code too. It is the right time to trim and rename it as it is a bit of mess right now. 
I plan to return the JWK related utility methods in the form of the proper API as well (such that we can have for example `sign(JWK jwk)`, plus a JWK set support, etc)
- Updated `JwtProvider` class to create `JwtProvider` once


sberyozkin:
@MikeEdgar thanks

PR #267, opened 2020-06-24
Allow for setting JWE headers after innerSign

### LINKED ISSUE #266: Impossible to set custom JWE headers after innerSign
Minor issue to do with `JwtSignature` `innerSign` methods returning `JwtEncryption` thus making it impossible to customize the JWE  encryption headers, spotted it while working on the test in the other branch. The fix is straightforward: return `JwtEncryptionBuilder` (which extends `JwtEncryption`) instead - this is totally backward compatible

sberyozkin:
Fixes #266


sberyozkin:
@MikeEdgar Thanks

PR #270, opened 2020-07-06
Update the version to 2.2.0.SNAPSHOT

sberyozkin:
Hi All, I'd like to move the version to `2.2.0` and release since `JWTParser` has had a new methods added and the decryption code has also been backported from `mpjwt12`. I suspect the lifespan of `2.2.x` won't be long given that `MP JWT 1.2` is hopefully is getting close and `mpjwt12` will set the version to `3.0.0`, but I thought it would be better to increase a minor version now anyway as there could be another couple of `2.2.x` releases in between....
I don't really mind much to do the next release as `2.1.3` since no breaking changes have been introduced but again I'll feel better when showing the release log to the Quarkus QE team if it becomes `2.2.0` :-)

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <groupId>io.smallrye</groupId>
     <artifactId>smallrye-jwt-parent</artifactId>
-    <version>2.1.3-SNAPSHOT</version>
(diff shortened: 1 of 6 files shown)


sberyozkin:
@radcortez Thanks

PR #271, opened 2020-07-07
Release Smallrye JWT 2.2.0

sberyozkin:


radcortez:
@sberyozkin just a question, this is out of date with the base branch. Is it ok?

sberyozkin:
Hi @radcortez I probably forgot to pull after I updated to `2.2.0-SNAPSHOT` before creating `release-2.2.0`, I think it should be safe as it only updates `` , but let me update the branch, would be safer

sberyozkin:
@radcortez Looks like it did not work:
https://github.com/smallrye/smallrye-jwt/runs/847394645

This is JavaDocs bug, I have to fix it, but I wonder why it was not detected here ?
Thanks

sberyozkin:
And how to ensure now that `2.2.0` release is done, not `2.2.1` ?

radcortez:
> @radcortez Looks like it did not work:
> https://github.com/smallrye/smallrye-jwt/runs/847394645
> 
> This is JavaDocs bug, I have to fix it, but I wonder why it was not detected as part of this PR ?
> Thanks

Yes, I guess we need to add javadoc build to make sure the release doesn't break.

> And how to ensure now that `2.2.0` release is done, not `2.2.1` ?

We need to revert the `project.yml` file.

PR #272, opened 2020-07-07
Revert "Release Smallrye JWT 2.2.0"

sberyozkin:
Reverts smallrye/smallrye-jwt#271


radcortez:
To revert, we need to push the `project.yml` directly without going through PR. Yes, not perfect, but there is no way to know if you want to do a release or a revert, unless you add some kind of metadata.

PR #276, opened 2020-07-08
Relax the key validation by default on MP JWT 1.2 branch

sberyozkin:
The verification key is now relaxed by default on the MP JWT 1.2 branch


sberyozkin:
Hi All, this PR goes into mpjwt12 and does not require a review, and all it does is it gets the TCK test passing by changing the relax key property from `false` to `true` (one of the options discussed earlier), so I'll go ahead with the merge, as I need to do one more MP JWT 1.2 sync on top of this merge

PR #296, opened 2020-08-17
Support custom schemes for the custom headers

### LINKED ISSUE #286: Scheme(s) should not be discared when alternate header is used
Is it intentional not to check for scheme(s) when alternate header is used or is it a bug?

When default  `Authorization` header is used, `smallrye.jwt.token.schemes` property is properly taken into account, but when alternate header name is used, for instance like this:
`smallrye.jwt.token.header=x-forwarded-authorization`
you're then completely discarding `smallrye.jwt.token.schemes` property.

Furthermore, you don't even allow a header value to contain any scheme.

For instance, if configured to use the default `Authorization` header, this header value is valid:
`Bearer <JWT token>`

But if `smallrye.jwt.token.header` property is set, then
`Bearer <JWT token>`
is not a valid header value.
Furthermore, any scheme name in front of a JWT itself is illegal, no matter if `smallrye.jwt.token.schemes` property is being used at all and no matter its value. Only this is valid header value if `smallrye.jwt.token.header` property is set:
`<JWT token>`

If this is a bug, it can be easily fixed by changing the `io.smallrye.jwt.auth.AbstractBearerTokenExtractor#getBearerToken` method here:
https://github.com/smallrye/smallrye-jwt/blob/4bdcdf758105ec96069287785218719bfb2f6c2a/implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java#L51

As you may guess, this is not an academic discussion: when Google Cloud Endpoints v2 is configured to use Firebase Auth, it forwards original `Authorization` header value as `x-forwarded-authorization` header value. So, as the original header value is forwared "as is", it contains the "Bearer " prefix. Hence the trouble...

sberyozkin:
Fixes #286

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/AbstractBearerTokenExtractor.java
@@ -65,11 +65,17 @@ public abstract class AbstractBearerTokenExtractor {
         } else if (AUTHORIZATION_HEADER.equals(tokenHeaderName)) {
             bearerValue = getBearerTokenAuthHeader();
         } else {
-            bearerValue = getHeaderValue(tokenHeaderName);
+            String customHeaderValue = getHeaderValue(tokenHeaderName);
 
-            if (bearerValue == null) {
+            if (customHeaderValue == null) {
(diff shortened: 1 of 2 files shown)


sberyozkin:
@MikeEdgar thanks

PR #298, opened 2020-08-21
Inject JWTCallerPrincipalFactory

### LINKED ISSUE #144: Allow unverified JWT
Hi,
I [asked in MP-JWT github](https://github.com/eclipse/microprofile-jwt-auth/issues/132) about a way to configure `no verification` for the JWT. It might be something proprietary so I'll try asking here too. Is it possible to inject an un-verified JWT? I had to manually parse (and verify the auth header) instead of just using `@Inject` like I think should be possible.

### LINKED ISSUE #193: Why is JWTCallerPrincipalFactory not injected into DefaultJWTParser

sberyozkin:
Fixes #144 
Fixes #193 

This is PR has been created to complete Andreas's effort (see #194) with a few minor differences. Here is the summary:
- `JWTCallerPrincipalFactory` is now injected into `DefaultJWTParser`
- I've added a producer and moved the serviceloader related code there - it is important to retain this option to avoid breaking the existing expectations around loading the factories
- TCK tests are passing, added 2 simple unit tests, updated the docs
- Added a [test in Quarkus](https://github.com/quarkusio/quarkus/compare/master...sberyozkin:smallrye-jwt-2.2.1?expand=1) as well

CC @andreas-eberle

### DIFF
--- a/doc/modules/ROOT/pages/configuration.adoc
+++ b/doc/modules/ROOT/pages/configuration.adoc
@@ -52,3 +52,43 @@ Smallrye JWT supports many properties which can be used to customize the token p
 |smallrye.jwt.verify.aud|none|Comma separated list of the audiences that a token `aud` claim may contain.
 |smallrye.jwt.required.claims|none|Comma separated list of the claims that a token must contain.
 |===
+
+= Custom Factories
+
+`io.smallrye.jwt.auth.principal.DefaultJWTCallerPrincipalFactory` is used by default to parse and verify JWT tokens and convert them to `JsonWebToken` principals.
+It uses `MP JWT` and `smallrye-jwt` properties listed in the `Configuration` section to verify and customize JWT tokens.
+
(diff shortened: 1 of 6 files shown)


sberyozkin:
@radcortez @MikeEdgar thanks

PR #302, opened 2020-08-25
Support X509Certificate token verification

### LINKED ISSUE #260: Support JWKs with the encoded `x5c` X509 certificate
A Quarkus user is asking about supporting `x5t` thumbprints, the initial step is to support the encoded `x5c` content to extract the public verification key out of it.

sberyozkin:
Fixes #260

This PR supports the verification of the signed tokens which contain either `x5t` or `x5t#S256` certificate thumbprints. If the key is JWK then it must contain an encoded `x5c` chain array with the 1st element of this array being the cert with the public key and whose thumbprint must match the one in the token.
Note we don't accept the `x5c` chain coming in the token itself as it would require the chain verification etc.
Also added 2 helper methods to `JwtSignatureBuilder` and spotted it contains the redundant `innerSign` declarations which are already present in its `JwtSignature` superclass


implementation/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -288,6 +288,19 @@ private static JWTAuthContextInfoProvider create(String publicKey, String public
...
+     * If this property is enabled then a signed token must contain either 'x5t' or 'x5t#256' X509Certificate thumbprint
+     * headers.
+     * Verification keys can only be in JWK or PEM Certificate key formats in this case.
+     * JWK keys must have a 'x5c' (Base64-encoded X509Certificate) property set.
+     * Note that 'smallrye.jwt.token.kid' property will have no effect as 'x5t' and 'x5t#S256'
+     * are the key identifiers when this property is set.
+     */
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.verify.certificateThumbprint", defaultValue = "false")
+    private boolean verifyCertificateThumbprint;
```
MikeEdgar, line 302: Does this need a line in `getOptionalContextInfo` to pass to the injected value to `JWTAuthContextInfo`?

sberyozkin: @MikeEdgar sorry, bad omission, in `DefaultJWTParserTest' the property is set in the test, so I added 2 more tests starting from [here](https://github.com/smallrye/smallrye-jwt/pull/302/files#diff-6547c7bfe75cb1091761931d682eccd4R124), have a look please.
There should be an easier way to bootstrap `JWTAuthContextInfoProvider` (may be just set all those properties as public, etc) but for now I'm keeping adding the helper methods

sberyozkin:
Hi @MikeEdgar, thanks

PR #303, opened 2020-08-25
Simpler token symmetric signing and encryption support

### LINKED ISSUE #290: Add JWT token sign/innerSign/encrypt methods accepting a secret key as String
We have `sign(String keyLocation)`/etc with the String parameter being the key location, but when the users need to secure a token using the symmetric algorithm like `HS256` they need to create `SecretKey` manually which is easy enough to do but it would be good if they can just do

`Jwt.upn("Alice").signWithSecret(generatedPassword);`
`Jwt.upn("Alice").innerSignWithSecret(generatedPassword).encrypt();`
`Jwt.upn("Alice").jwe().encryptWithSecret(generatedPassword);`

sberyozkin:
Fixes #290. 

Quarkus users have to manually create `SecretKey` when they want to sign the token with a password so this PR adds a few shortcuts making it very easy to do.

### DIFF
--- a/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
+++ b/implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
@@ -27,6 +27,7 @@ import javax.inject.Inject;
 
 import org.eclipse.microprofile.jwt.JsonWebToken;
 
+import io.smallrye.jwt.KeyUtils;
 import io.smallrye.jwt.algorithm.KeyEncryptionAlgorithm;
 import io.smallrye.jwt.algorithm.SignatureAlgorithm;
 import io.smallrye.jwt.auth.cdi.JWTCallerPrincipalFactoryProducer;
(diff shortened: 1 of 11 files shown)


implementation/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
```diff
@@ -130,4 +142,9 @@ private void setKeyEncryptionAlgorithmIfNeeded(JWTAuthContextInfo newAuthContext
             newAuthContextInfo.setKeyEncryptionAlgorithm(newAlgo);
         }
     }
+
+    private SecretKey createSecretKeyFromSecret(String secret) {
+        byte[] secretBytes = secret.getBytes(StandardCharsets.UTF_8);
+        return new SecretKeySpec(secretBytes, "AES");
+    }
```
MikeEdgar, line 149: Duplicate of `JwtBuildUtils.createSecretKeyFromSecret` (FYI)

sberyozkin: Hi Mike @MikeEdgar thanks for spotting it, yes, I recall thinking about, I wanted to keep the `build` related code  as isolated as possible in case we consider in time moving it in its own module, but now I've realized this code really belongs in `KeyUtils` :-). So I've just moved it there, and will merge shortly, cheers

PR #305, opened 2020-08-26
Support for the configured secret key location

### LINKED ISSUE #126: Support the symmetric signature verification

sberyozkin:
Fixes #126 
This is the final PR before the planned 2.2.1 release and resolves a long pending issue to do with configuring the symmetric key locations, with some good user interest around it. Now we'll have a comprehensive support for securing and validating JWT tokens with all sort of keys which will work well in a variety of cases.
PR itself is basic, I had to add a new property to avoid overloading the existing `publickey.location` one.

### DIFF
--- a/doc/modules/ROOT/pages/configuration.adoc
+++ b/doc/modules/ROOT/pages/configuration.adoc
@@ -31,6 +31,7 @@ Smallrye JWT supports many properties which can be used to customize the token p
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
+|smallrye.jwt.verify.key.location|NONE|Location of the verification key which can point to both public and secret keys. Secret keys can only be in the JWK format. Note that 'mp.jwt.verify.publickey.location' will be ignored if this property is set.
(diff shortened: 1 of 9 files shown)


sberyozkin:
Hi @MikeEdgar thanks, I did a minor update in `JWTAuthContextInfoProvider` (the docs were between `Inject` and `ConfigProperty` while for all other properties - before `Inject` so I just moved it above `Inject`) and yesterday, after I signed off, I realized it would be good to update the unit tests to check the secret key was loaded, so I did it as well. 
Will merge soon, thanks

PR #307, opened 2020-08-28
Minor JavaDoc updates

sberyozkin:


sberyozkin:
Hi @MikeEdgar This is just a minor JavaDoc updates as no `secret` param was available in the new builder methods, so I'll just merge it :-)

PR #329, opened 2020-11-17
Update the build and release workflows to setup java at v1

sberyozkin:


radcortez (changes requested):
The version also needs to be updated in the release flow. Thanks!

radcortez (approved):

Issue #327, opened 2020-11-10
JWT builder NPE against the key generated with keytool


sberyozkin:
See https://stackoverflow.com/questions/63971028/quarkus-mp-jwt-how-to-generate-the-key-pair

Issue #321, opened 2020-10-15
Use JWK `kid` when generating the tokens if no `kid` is already set


sberyozkin:

sberyozkin:
I've spent quite a bit of time on it, and had a fix, but  then realized them same should apply to the algorithm property which JWK is expected to contain so I've decided not to rush with this trying to fix this issue for 2.4.0

sberyozkin:
It should work with a single JWK - when both `kid` and `alg` (and `encAlg` for encryption) are collected from JWK - and with a multi-key JWK set - where `kid` is already known but `alg`/`encAlg` - may not yet

sberyozkin:
I'll try to fix this issue asap and release 3.1.1 (after updating the forum)

PR #331, opened 2020-11-23
Split the implementation module into common, jwt-auth, jwt-build

sberyozkin:
This PR simply splits the current `implementation` module as per the smallry forum discussion, 3 modules are introduced:
- `common` (`smallrye-jwt-common` artifact id) module: where `KeyUtils` and `ResourceUtils` utility code has been moved to
- `jwt-build` (`smallrye-jwt-build` artifact id) module: where the JWT build related code has been moved to
- `jwt-auth` module: the current `implementation` module - except that the utility and build code is now coming from `smallrye-jwt-common`and `smallrye-jwt-build`; I think we can drop the `smallrye-jwt-build` dependency for `3.0.0` and let users who need it include it.

-  `KeyUtils` and `ResourceUtils` are in `io.smallrye.jwt` package on master and in `io.smallrye.jwt.util` package in `common` so I've kept the original classes in `jwt-auth` for now but deprecated them and updated to delegate to their `common` friends. These `jwt-auth` deprecated classes can be dropped in `3.0.0`

This is really it - plus some minor dependency clean-up made possible by this split.

After this PR is reviewed/merged I will follow up with 2 small fixes (#327, #321) and release `2.4.0`


MikeEdgar (approved):
Looks good, I didn't notice anything outside the described changes.

radcortez:
LGTM

radcortez (approved):

sberyozkin:
@MikeEdgar @radcortez thanks, that should improve the modularization picture; for 3.0.0 I'll propose adding `jwt-jaxrs` and move the jaxrs specific code into it - right now, for example, in Quarkus we do not use the JAX-RS classes

sberyozkin:
@MikeEdgar by the way, I meant to invite you to MP JWT gitter channel but ended up adding you, I thought it would send an invitation but looks like you are already in that room :-), sorry about me somehow skipping the invitation step; please don't hesitate to `pause the memebership` :-) of that room/channel if it becomes too noisy - don't think it will be noisy there though... thanks

MikeEdgar:
@sberyozkin - no problem. I realize now that I have not been getting notifications for that particular channel. I'll need to look into that :-)

PR #334, opened 2020-11-25
Better handling of the wrong JWT build key locations

### LINKED ISSUE #327: JWT builder NPE against the key generated with keytool

sberyozkin:
Fixes #327 

The NPE reported at the StackOverflow page is to do with the key being null, so this PR ensures it won't happen again and makes it consistent between the signature and encryption implementations.

Also the key format in that issue is PKCS12 which is not currently supported - perhaps we can do it in 3.0.0

I'd like to release 2.4.0 after this PR is reviewed/merged


radcortez (approved):

sberyozkin:
@radcortez thanks

PR #349, opened 2020-12-11
Use UUID for the 'jti' claim, and update ResourceUtils to have a separate InputStream utility

### LINKED ISSUE #344: Use a crypto-strong method for generating `jti`
Right now it is generated for the new tokens without SecureRandom so `SecureRandom` or just `UUID` should be used instead.
Without it a possibility of 2 different tokens having the same `jti` is still small but it should be much smaller.

sberyozkin:
Fixes #344 

Also, minor refactoring of ResourceUtils to be able to use the utility code which gets the stream only, without converting it to String, realized it can be handy while I was coding some KeyStore.load

### DIFF
--- a/implementation/common/src/main/java/io/smallrye/jwt/util/ResourceUtils.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/util/ResourceUtils.java
@@ -41,6 +41,28 @@ public final class ResourceUtils {
 
     public static String readResource(String resourceLocation, UrlStreamResolver urlResolver) throws IOException {
 
+        InputStream is = getResourceStream(resourceLocation, urlResolver);
+
+        if (is == null) {
+            return null;
+        }
+
+        StringWriter contents = new StringWriter();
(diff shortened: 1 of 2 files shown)


radcortez (approved):
We had  some issues in SR Config and Quarkus with InputStream locking the jar. I'm not sure if this is the case, but I'm going to add a runner for Windows here so we can test for that issue as well.

radcortez:
BTW, before we merge something we should decide if we want to merge the `mpjwt12` branch to master?

sberyozkin:
@radcortez Sure, I'm just opening the PRs today, but won't merge them before next week, I was thinking getting these PRs into 3.0.0.RC2 and do 2.4.1 as well early next week,  we probably want to keep the master branch on 2.4.x untill we get the final 3.0.0 release ?
The 3 PRs I've opened are isolated to the builder API

radcortez:
Sounds good.

MikeEdgar (dismissed):

sberyozkin:
@radcortez @MikeEdgar thanks  I've just rebased this branch so if it goes green then I'll merge

PR #351, opened 2020-12-11
Support for expireIn build method and other minor cleanup

sberyozkin:
This PR adds `expiresIn` helper to set the `exp` claim which should make it much easier to set the custom expiry claim as opposed to the users having to manually add the lifespan to the current issued time. Plus some minor cleanup

### DIFF
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/JwtClaimsBuilder.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/JwtClaimsBuilder.java
@@ -1,5 +1,6 @@
 package io.smallrye.jwt.build;
 
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


radcortez (dismissed):

implementation/jwt-build/src/main/java/io/smallrye/jwt/build/JwtClaimsBuilder.java
```diff
@@ -81,19 +82,37 @@ default JwtClaimsBuilder issuedAt(Instant issuedAt) {
...
+    JwtClaimsBuilder expiresIn(long expiresIn);
+
+    /**
+     * Set an expiry 'exp' claim
+     * 
+     * @param expiresIn the relative expiry duration which will be converted to seconds and added to the issuedAt time
+     * @return JwtClaimsBuilder
+     */
+    default JwtClaimsBuilder expiresIn(Duration expiresIn) {
+        return expiresAt(expiresIn.getSeconds());
```
MikeEdgar, line 115: Should this be delegating to `expiresIn(long)` instead?

MikeEdgar (changes requested):
Looks good, but I think there's one minor typo.

implementation/jwt-build/src/main/java/io/smallrye/jwt/build/JwtClaimsBuilder.java
sberyozkin: @MikeEdgar sorry :-) fixed and added more tests, pushing shortly

sberyozkin: That is done, thanks Mike for spotting it

MikeEdgar (approved):

PR #354, opened 2020-12-15
Revert "Bump smallrye-parent from 23 to 25"

sberyozkin:
Reverts smallrye/smallrye-jwt#346


sberyozkin (commented):
I think this one may've caused a release failure

PR #359, opened 2020-12-15
Fix JaCoCo Test Coverage.

radcortez:


sberyozkin (approved):
Hey Roberto, that does look good (thanks for some cleanup in the commons too) lets wait till tomorrow morning for Mike to have a look as well as I think he was involved somehow - Mike if you have a few mins - please check, ready to go from my point of view

MikeEdgar:
I think I finally figured this out in OpenAPI. My understanding is that you need either a `sonar.coverage.jacoco.xmlReportPaths` property in each module with a **relative path** to the aggregated report -or- a single top-level `sonar.coverage.jacoco.xmlReportPaths` property that contains the relative values for all of the modules (see OpenAPI). +1 to having a separate `coverage` module, I think we should do that elsewhere also.

https://github.com/smallrye/smallrye-open-api/blob/4d5a8f414264cae48d0f79c2740e708a545ff201/pom.xml#L35-L46

radcortez:
Yes, and I've found a way to simplify it. I'll update.

radcortez:
Ok, so yes. A `sonar.coverage.jacoco.xmlReportPaths` with multiple paths or have each module pointing to the same aggregate one works.

I've mostly followed the recommend approach:
https://community.sonarsource.com/t/coverage-test-data-importing-jacoco-coverage-report-in-xml-format/12151
https://github.com/SonarSource/sonar-scanning-examples/tree/master/sonarqube-scanner-maven/maven-multimodule

A trick is to set `${maven.multiModuleProjectDirectory}/coverage/target/site/jacoco-aggregate/jacoco.xml`, which will always point to the right coverage file in every module and have a profile in parent to set it across the project.

MikeEdgar:
I hadn't used `maven.multiModuleProjectDirectory` before, but that looks like a nice approach.

MikeEdgar (approved):

PR #367, opened 2020-12-24
Move the JAX-RS related code into smallrye-jwt-jaxrs

### LINKED ISSUE #332: Improve jwt-auth modularization before 3.0.0 release

sberyozkin:
Fixes #332

This PR tries to improve the modularization further and moves the JAX-RS related code into its own module `smallrye-jwt-jaxrs` - for example, in Quarkus, the JAX-RS part is not used.
My only tiny concern here is that in `SmallRyeJWTAuthCDIExtension` the JAX-RS filter is now loaded reflectively - this is fine for the TCK tests though.
This PR targets the `mpjwt12` branch.
Lets check it early next year :-)


MikeEdgar (approved):

sberyozkin:
@MikeEdgar thanks, and happy holidays; I'll also wait for Roberto and especially for Darran to review as he may be using some of that code, cheers

radcortez:
> My only tiny concern here is that in `SmallRyeJWTAuthCDIExtension` the JAX-RS filter is now loaded reflectively - this is fine for the TCK tests though.

We can probably mode the CDI Extension to its own module as well: `jwt-cdi`. I've done that with SR Config. Quarkus does not need the extension.

radcortez (approved):

sberyozkin:
@radcortez Yeah, that can be further optimized. I'll ping Darran now to check too

darranl (approved):
Adding an approval, we don't use the JAX-RS integration ourselves within WildFly so will be good to separate.

sberyozkin:
Thanks all

PR #371, opened 2021-01-06
Corrected toString implementation

bobbyphilip:
Corrected toString implementation when the groups are empty


sberyozkin:
@bobbyphilip Thanks for the PR, can you please rebuild locally - there are some formatting issues. By the way, maybe we can just drop that `setLength` call completely ?

sberyozkin (approved):

PR #373, opened 2021-01-08
Issue#372 Status code should be 500 when public key cannot be resolved

### LINKED ISSUE #372: Error code should be 500 when a public key cannot be extracted from the mp.jwt.verify.publickey.location property
Hi all, 

I am currently working on resolving the WildFly issue [WFLY-13164](https://issues.redhat.com/browse/WFLY-13164). I noticed that `JWTHttpAuthenticationMechanism` sends the 401 status code when validating a request throws an exception.

However, when this exception is caused by failing to extract a public key from the `mp.jwt.verify.publickey.location` property (see `DefaultJwtTokenParser` and `KeyLocationResolver`), I was wondering whether it would be more appropriate to send a 500 status code instead.

Perhaps we can throw a 500 status code when we catch a `ParseException`? 

Thanks for your help, 
Sonia

SoniaZaldana:
Fixes #372 

Currently, exception handling in `JWTHTTPAuthenticationMechanism` issues status code 401 when the public key is corrupted or unavailable.

### DIFF
--- a/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/mechanism/JWTHttpAuthenticationMechanism.java
+++ b/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/mechanism/JWTHttpAuthenticationMechanism.java
@@ -15,6 +15,7 @@
  */
 package io.smallrye.jwt.auth.mechanism;
 
+import java.io.IOException;
 import java.util.Set;
 
 import javax.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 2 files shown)


sberyozkin (approved):
@SoniaZaldana, great, thank you.

I'd just like to wait till later today and see if Mike has the time to approve, he has worked on this mechanism originally

Thank you

MikeEdgar:
@SoniaZaldana , @sberyozkin - something I noticed (that may not be an issue), is that if an `UnresolvableKeyException` occurs in `DefaultJWTTokenParser#decryptSignedToken`, the `ParseException` will not include it as the cause. Also, if somehow an unexpected runtime exception is thrown, it might make sense to signal the 500 error as well.

sberyozkin:
@MikeEdgar Hi, thanks for the careful review as always - I agree it makes to fix for 2.4.2. 
Let me merge it though as @SoniaZaldana is looking at a blocker issue :-), and I take care of your comment before 2.4.2 in the next couple of days - I'll open a new issue immediately

PR #375, opened 2021-01-11
Report UnresolvableKeyException in the inner-sign/decrypt case, update JAX-RS filter too

sberyozkin:
@MikeEdgar - have a look please - I've also updated the JAX-RS filter 

@SoniaZaldana FYI, I also had to update the `MechanismLogging.noUsableKey` log message a tiny bit, dropped `...public...` in the key description - it can also be a private decryption key (standard from MP JWT 1.2) or a symmetric key (supported at the smallrye-jwt level only). I think we can improve it further down the line with the key type specific exceptions as well...
You are also welcome to review (I'll be able to request a review when you become a committer )

By the way, for 3.0.0, I think moving the http mechanism into its own module would be the right idea :-) as the core module would better be transport agnostic; similarly to how the jaxrs code got moved into its own module

Thanks


radcortez (approved):

SoniaZaldana (approved):

MikeEdgar (approved):
Looks good!

sberyozkin:
Thanks all

PR #376, opened 2021-01-12
Add a smallrye.jwt.new-token.audience property

sberyozkin:
Fixes #364

This PR adds a new `smallrye.jwt.new-token.audience` property - for the users be able to avoid setting the `technical` claims like `aud` (similarly to how it has already been done with `iss`)

PR #377, opened 2021-01-12
Add PrincipalUtils

### LINKED ISSUE #362: Allow to disable token validation
In the context of a Quarkus application I would like to use the `quarkus-smallrye-jwt` extension and thus `smallrye-jwt` to set the Quarkus `SecurityIdentity` based on a JWT token passed as a HTTP header. However, since the JWT token will already have been validated (in our case by Azure API Management) I would like to disable the validation performed by `smallrye-jwt`, while still have the Quarkus extension set the `SecurityIdentity`.

I couldn't find any config property which would allow me to completely disable the validation, which is why I report this issue.

sberyozkin:
Fixes #362

This PR only moves the code which updates the claims based on some smallrye-jwt specific properties as well as sets a raw token.
This is to help the users who write the custom factories in order to skip the token validation but still wish to apply these `post-verification` properties.

I'd like to release `2.4.2` once #376 and this PR get merged

### DIFF
--- a/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
+++ b/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
@@ -18,13 +18,7 @@ package io.smallrye.jwt.auth.principal;
 
 import static java.util.Collections.emptyList;
 
-import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


radcortez (approved):

sberyozkin:
Mike, merging this one too - this is just a straightforward move of the code which only updates the claims for the custom factories to reuse it; ping me please if you'd like to clarify something

PR #378, opened 2021-01-12
Add smallrye-jwt-http-mechanism and smallrye-jwt-cdi-extension modules

### LINKED ISSUE #332: Improve jwt-auth modularization before 3.0.0 release

sberyozkin:
Fixes #332

This PR targets the `mpjwt12`branch and moves `JwtHttpAuthenticationMechanism` into its own `smallrye-jwt-http-mechanism` module and `SmallRyeJWTAuthCDIExtension` into its own `smallrye-jwt-cdi-extension` (the latter can now include an extension provider resource as well - I did not do it in this PR to avoid affecting TCK somehow due it being extended in the TCK)

Also CC @SoniaZaldana


radcortez (approved):

sberyozkin:
Hi @MikeEdgar @darranl I'm going to postpone merging this PR for another day in case you'd like to comment - in meantime I'll release 2.4.2 - the plan is to try merge this PR and do 3.0.0.RC2 tomorrow; 
Mike - in the JAX-RS feature (`smallrye-jwt-jaxrs`) - I removed a check where the HTTP mechanism availability is asserted since the CDI extension is now in its own module - so I assumed it would be reasonable to expect that it would now become an issue of the dependency management by the user. Otherwise the priority is always given to the HTTP mechanism which is not what the users of `smallrye-jwt-jaxrs` may want; I guess that is the only 'complex' part of this PR :-) and I guess we can fine tune it going forward as well.
Thanks

MikeEdgar:
@sberyozkin - I'm a little confused about the source/target branches for the PR. It seems some of the commits on the PR branch are already in `mpjwt12`.

radcortez:
It seems the `mpjwt12` was rebased and now the source for this PR also needs rebase.

sberyozkin:
Sorry, will do it

sberyozkin:
@radcortez @MikeEdgar It can cheaper/faster for me to create a new PR - this actually shows the danger of rebasing on a topic branch, something @darranl implied - I was so happy the earlier master commits have been rebased cleanly, except that I had another PR against the same branch open...

PR #379, opened 2021-01-12
Remove Optional for the config properties with the default values

sberyozkin:
This is the last PR which I'd like to target at the `mpjwt12` branch. It updates `JwtAuthContextInfoProvider` and:
- removes `Optional` where possible (Mike - thanks for pointing it earlier on)
- removes the redundant getters - they do not even cover all the properties - they were originally intended for Thorntail and even there there were not property integrated as far as I recall
- drops the deprecated property which has already been deprecated in the `2.x` line


radcortez (approved):

sberyozkin:
This is just some cleanup so going ahead with Roberto approving

PR #381, opened 2021-01-13
Create smallrye-jwt-http-mechanism and smallrye-jwt-cdi-extension modules

### LINKED ISSUE #332: Improve jwt-auth modularization before 3.0.0 release

sberyozkin:
Hi All, here is a new PR, thanks

Fixes #332


implementation/jwt-jaxrs/src/main/java/io/smallrye/jwt/auth/jaxrs/SmallRyeJWTAuthJaxRsFeature.java
```diff
@@ -48,11 +46,7 @@ public boolean configure(FeatureContext context) {
         if (enabled) {
             context.register(JWTAuthorizationFilterRegistrar.class);
 
-            if (!SmallRyeJWTAuthCDIExtension.isHttpAuthMechanismEnabled()) {
-                context.register(JWTAuthenticationFilter.class);
-
-                JAXRSLogging.log.eeSecurityNotInUseButRegistered(JWTAuthenticationFilter.class.getSimpleName());
-            }
+            context.register(JWTAuthenticationFilter.class);
```
MikeEdgar, line 49: @sberyozkin - I'm trying to get my head around the impact of registering the filter unconditionally. Suppose the HttpAuthMechanism is registered via the CDI extension. I assume (but I am not certain) that the filter would then execute second, and the principal would be an instance of `JsonWebToken`, bypassing the code in `filter`. What are your thoughts?

MikeEdgar (approved):
Just a question on the JAX-RS change, but otherwise it looks good to me. Thanks @sberyozkin .

implementation/jwt-cdi-extension/pom.xml
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+ ~  Copyright 2017 Red Hat, Inc.
```
darranl, line 4: As a new file should this be 2021?

implementation/jwt-http-mechanism/pom.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+ ~  Copyright 2017 Red Hat, Inc.
```
darranl, line 4: Also 2021?

implementation/jwt-jaxrs/src/main/java/io/smallrye/jwt/auth/jaxrs/SmallRyeJWTAuthJaxRsFeature.java
sberyozkin: Hi @MikeEdgar I think it would be simplest to copy that code from the feature checking for `HttpAuthMechanism` just to preserve the original logic. We can fine tune it further as required... Otherwise, indeed, I'm not 100% sure of the side-effects introduced by this update - it does not look like this feature is used right now, but it is better to keep the check in place

darranl:
@sberyozkin my day is a little bit disrupted today but do you want me to give this a go with WildFly and double check no unintended consequences?  Otherwise looks good to me as well.

sberyozkin:
@darranl Thanks, I update the year in a few poms. As discussed, we can do RC2 - it also makes sense to try in Quarkus before the final, we can do RC2 later today

implementation/jwt-jaxrs/src/main/java/io/smallrye/jwt/auth/jaxrs/SmallRyeJWTAuthJaxRsFeature.java
sberyozkin: @MikeEdgar I've tried to do this copy but found it was not easy to do. I suppose what I said in the closed PR probably makes sense. The user should be now responsible for making sure no ambiguities happen.
This JAX-RS feature is not annotated as `Provider` and it is not loaded in the CDI extension. So if some users would like to use this feature - they'd have to enable it in their `Application` or via `DynamicFeature`.
And if the users choose to use the CDI extension then all is good - it will load the HTTP mechanism first if EESec is enabled and if not - it will register the JAX-RS filter directly.
So I think all is good :-) - do you agree ?

MikeEdgar: @sberyozkin - I agree. Also, for historical perspective, much of this code was originally intended to allow using SmallRye JWT in Wildfly prior to it's official inclusion.

sberyozkin: @MikeEdgar OK, thanks. Ok then I'll go ahead with the merge now that Darran also checked it for us to do RC2 today

PR #392, opened 2021-01-21
Update MP Config version to 2.0

### LINKED ISSUE #337: Update MP Config dependency to 2.0.
It requires a RESTEasy release (4.6.0) to run the TCK, because MP Config 2.0 has breaking changes that affect RESTEasy.

sberyozkin:
Fixes #337

@radcortez I'm seeing some strange TCK dep issue with 4.6.0.Final - hoping it is a problem with the cached staged repo

### DIFF
--- a/implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtBuildConfigSource.java
+++ b/implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtBuildConfigSource.java
@@ -1,7 +1,10 @@
 package io.smallrye.jwt.build;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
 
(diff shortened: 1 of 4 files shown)


sberyozkin:
Oh yea, I could not find it, the staged repo was in the root pom, but it makes no difference

radcortez:
This update is not going to work because the dependencies from 3.x to 4.x changed a bit. For instance the missing dependency was discontinued in 4.x

Also, this is tied to the Arquillian Wildfly we use to run tests, which does not support RESTEasy 4.x. I think we need to move this to use Jetty and add the Weld and RESTEasy, so we can build our environment with the versions we want. I've done this for OpenAPI, GraphQL and OpenTracing and seems to work. I can send a PR for this later.

sberyozkin:
OK, I've sorted it out

sberyozkin:
@radcortez sorry - missed your comment - I used `resteasy-undertow`

sberyozkin:
Looks like it worked :-)

sberyozkin:
May be because TCK has no `ConfigSources` ?

radcortez:
Hum.... maybe. Still, I think we should move to Jetty with custom dependencies with Weld, RESTEasy and Yasson, so we don't have a circular dependency between us and Wildly (which in most cases is fine, but we already had some issues in other impls).

radcortez:
Sure, lets move forward as is and I'll update it later. Thanks!

radcortez (approved):

sberyozkin:
Hey @radcortez grand :-), lets do it

PR #394, opened 2021-01-22
Fix the default algorithm typo

sberyozkin:


sberyozkin:
@radcortez @MikeEdgar this is a typo I did with the last fix - this code is not used anywhere in smallrye-jwt - but it already made it to 3.0.0-SNAPSHOT which was detected on the Quarkus MP 4.0 branch in a test not related to MP JWT.
I'll merge right now

sberyozkin:
CC @darranl FYI - but as I said - it is not used anywhere in smallrye-jwt - I just kept that method mechanically but made a typo...

PR #400, opened 2021-01-26
Add methods for secret key generation (#342)

### LINKED ISSUE #342: Add a method for generating a correct SecretKey
A good number of projects (Jose4J itself, [JJWT](https://github.com/jwtk/jjwt#jws-key-create) and others) have the utility code for generating the keys. Jose4J itself will enforce that the key length meets the algorithm constraints so the users are not able to use the wrong key size, and Jose4J would report the required length, but having a correct key upfront would be useful.

This feature is not that important for the asymmetric algorithms, in the production one would expect the keys already distributed or available.

This is mainly of interest for generating the correct `SecretKey` - in scope of creating a JWT cookie, as again, one would expect the key be available on both ends.

[KeyUtils.generateSecretKey](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java#L244) already exists, but takes a String. So we need another one taking a signature or key encryption algorithm which will return the right key length. `Jose4j` may already have it so we will just wrap around it

ejba:
Closes #342

### DIFF
--- a/implementation/common/pom.xml
+++ b/implementation/common/pom.xml
@@ -51,6 +51,11 @@
       <groupId>org.jboss.logging</groupId>
       <artifactId>jboss-logging-processor</artifactId>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
 </project>
--- a/implementation/common/src/main/java/io/smallrye/jwt/util/JWTUtilMessages.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/util/JWTUtilMessages.java
@@ -2,6 +2,7 @@ package io.smallrye.jwt.util;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.NoSuchAlgorithmException;
 
 import org.jboss.logging.Messages;
@@ -21,4 +22,7 @@ interface JWTUtilMessages {
 
     @Message(id = 2, value = "No resource with the named %s location exists")
     IOException keyNotFound(String keyLocation);
+
+    @Message(id = 3, value = "Algorithm %s is not a symmetric-key algorithm")
+    InvalidAlgorithmParameterException requiresSymmetricAlgo(String algorithmName);
 }
--- a/implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
@@ -24,6 +24,7 @@ import java.io.StringReader;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.Key;
 import java.security.KeyFactory;
 import java.security.KeyPair;
@@ -31,6 +32,7 @@ import java.security.KeyPairGenerator;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
+import java.security.SecureRandom;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.security.spec.PKCS8EncodedKeySpec;
@@ -39,6 +41,7 @@ import java.security.spec.X509EncodedKeySpec;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collections;
+import java.util.EnumMap;
 import java.util.List;
 
 import javax.crypto.SecretKey;
@@ -66,6 +69,22 @@ public final class KeyUtils {
     private static final String EC = "EC";
 
     private KeyUtils() {
+    }
+
+    protected static final EnumMap<KeyEncryptionAlgorithm, Integer> KEY_ENCRYPTION_BITS = new EnumMap(
+            KeyEncryptionAlgorithm.class);
+    static {
+        KEY_ENCRYPTION_BITS.put(KeyEncryptionAlgorithm.A128KW, 128);
+        KEY_ENCRYPTION_BITS.put(KeyEncryptionAlgorithm.A192KW, 192);
+        KEY_ENCRYPTION_BITS.put(KeyEncryptionAlgorithm.A256KW, 256);
+    }
+
+    protected static final EnumMap<SignatureAlgorithm, Integer> SIGNATURE_ALGORITHM_BITS = new EnumMap(
+            SignatureAlgorithm.class);
+    static {
+        SIGNATURE_ALGORITHM_BITS.put(SignatureAlgorithm.HS256, 256);
+        SIGNATURE_ALGORITHM_BITS.put(SignatureAlgorithm.HS384, 384);
+        SIGNATURE_ALGORITHM_BITS.put(SignatureAlgorithm.HS512, 512);
     }
 
     public static PrivateKey readPrivateKey(String pemResName) throws IOException, GeneralSecurityException {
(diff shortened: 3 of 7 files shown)


sberyozkin:
Hi @ejba thanks for giving a try :-), this PR goes in the right direction IMHO, but I've been thinking about reusing the existing enums.
So both `SignatureAlgorithm` and `KeyEncryptionAlgorithm` have algorithm enums which require the secret keys, so the new enum duplicates some `SignatureAlgorithm` enums and also misses on some `KeyEncryptionAlgorithm` enums.

Adding Assymmetric and Symmetric variations would make 4 classes here - possibly cleaner - but this is also raises the question of the API and configuration support.

Can we get the key size property added to the existing enums ?  Plus may be we can also have a standalone enum, `AlgorithmType` (`ASSYMMETRIC`, `SYMMETRIC`) with each enums returning the type. Then we'd have methods `createSecretKey(SignatureAlgorithm algo)`, `createSecretKey(KeyEncryptionAlgorithm algo)`, these would throw IAE if the algo is not symmetric. Later we can also do something for the assymetric keys.

Does it sound reasonable to you ?

sberyozkin:
@ejba It is also better reuse the code, inside the method which converts a secret string to `SecretKey` as opposed to using the key factory, as it will not know of some algorithms and we don't need to worry about converting to the Java equivalents - this code uses `AES` scheme (similar code exists in Jose4j) which is fine even when we need the keys for `HS` - all we really need there is the key size :-)

implementation/common/pom.xml
```diff
@@ -51,6 +51,11 @@
       <groupId>org.jboss.logging</groupId>
       <artifactId>jboss-logging-processor</artifactId>
     </dependency>
+    <dependency>
```
sberyozkin, line 54: I've noticed we use 
```
<dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <scope>test</scope>
    </dependency>
```
in other modules - I don't mind introducing a jupiter dependency, but I guess it would make sense to have it consistent as part of the dedicated PR ? I suppose we can keep this dependency for this PR and then follow up...

sberyozkin:
@ejba Hi, thanks very much for spending the time on it.

I have to apologize. While I do like the idea of setting the key sizes property, I think my proposal does not add much clarity and was rushed. For example, in case of `RS256` the actual key is required to be 2048bits+, etc for some other signature and encryption algorithms (inlc PBEs - everything is derived from the user password). Sorry for asking to follow it - I'm not really happy with it now that I see how it looks like :-)

I think going forward some deeper thinking is required how to make it clean and obvious, may be taking your original idea as an input where we have asymmetric and symmetric signature algorithms.

So I don't know what is the best option for now.

Perhaps, for now, we should only have a single method:

`generateSecretKey(SignatureAlgorithh algorithm)` ?

If it is not one of the `HS` algorithms - IAE, else - we know the key size from the algorithm enum name.

same for

`generateSecretKey(KeyEncryptionAlgorithh algorithm)` ?

(lets only check `A***KW` algorithm - the length is known from the name)

both methods will delegate to the internal method generating a key (not sure we need a public one accepting `byte[]`)

Can you please do another iteration :-) ?

ejba:
Done :)

sberyozkin:
Hi Emanuel, thanks, let me comment please

implementation/common/src/main/java/io/smallrye/jwt/algorithm/SymmetricKeyAlgorithm.java
```diff
@@ -0,0 +1,27 @@
+package io.smallrye.jwt.algorithm;
```
sberyozkin, line 1: Can you please remove this file for now

implementation/common/src/main/java/io/smallrye/jwt/util/JWTUtilMessages.java
```diff
@@ -2,6 +2,7 @@
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.security.InvalidAlgorithmParameterException;
```
sberyozkin, line 5: IMHO it should be `IllegalArgumentException` - it is all about the correct algo parameter parameter - it is not happening at the processing time

sberyozkin: Ignore it please - this is actually a prefect exception type for this case :-)

implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
```diff
@@ -246,6 +249,42 @@ public static SecretKey createSecretKeyFromSecret(String secret) {
...
 
+    /**
+     * Generates a SecretKey.
+     *
+     * @param algo key encryption algorithm.
+     * @return SecretKey.
+     * @throws NoSuchAlgorithmException algorithm not found.
+     */
+    public static SecretKey generateSecretKey(KeyEncryptionAlgorithm algo) throws InvalidAlgorithmParameterException {
+        final String algoName = getSymmetricAlgoName(algo);
```
sberyozkin, line 260: Here we only need to check if it is one of 3 `KW` algorithms and get the bits from the name

implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
```diff
@@ -246,6 +249,42 @@ public static SecretKey createSecretKeyFromSecret(String secret) {
...
+
+        int keySizeBits = getKeySizeFromAlgo(algo);
```
sberyozkin, line 262: Please have an internal function which will use `AES` scheme - that will be fine - since it is not really used

implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
sberyozkin: You can safely drop `getSymmetricAlgoName` for now - even if it works within the `encryption` variation

implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
```diff
@@ -273,6 +312,38 @@ public static PublicKey decodeEncryptionPublicKey(String pemEncoded, KeyEncrypti
         return kf.generatePublic(spec);
     }
 
+    static String getSymmetricAlgoName(KeyEncryptionAlgorithm algo) throws InvalidAlgorithmParameterException {
```
sberyozkin, line 315: Please drop this method

implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
```diff
@@ -273,6 +312,38 @@ public static PublicKey decodeEncryptionPublicKey(String pemEncoded, KeyEncrypti
...
+        if (algoName.startsWith("HS")) {
+            String extractedSize = algoName.substring(2);
+            return Integer.parseInt(extractedSize) / 8;
+        }
+        throw JWTUtilMessages.msg.requiresSymmetricAlgo(algo.name());
+    }
+
+    static int getKeySizeFromAlgo(KeyEncryptionAlgorithm algo) throws InvalidAlgorithmParameterException {
+        String algoName = algo.name();
+        if (algoName.startsWith("PBES2_HS")) {
```
sberyozkin, line 340: Please only check KW algorithms, https://github.com/smallrye/smallrye-jwt/blob/master/implementation/common/src/main/java/io/smallrye/jwt/algorithm/KeyEncryptionAlgorithm.java#L15

implementation/common/src/test/java/io/smallrye/jwt/util/GenerateSecretFromSignatureAlgoTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.runners.Parameterized;
+
+import io.smallrye.jwt.algorithm.SignatureAlgorithm;
+
+@RunWith(Parameterized.class)
+public class GenerateSecretFromSignatureAlgoTest {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
```
sberyozkin, line 22: I thin we can have a nice internal map like this inside `KeyUtils` - where we can can get the key bits from the map given an enum value, and also we can get another map like "HS256:HMAC", etc for remaining `HS` and also `A...KW` algos (`AES`) scheme

implementation/common/src/test/java/io/smallrye/jwt/util/GenerateSecretFromKeyEncryptionAlgoTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import io.smallrye.jwt.algorithm.KeyEncryptionAlgorithm;
+
+@RunWith(Parameterized.class)
+public class GenerateSecretFromKeyEncryptionAlgoTest {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+                { "PBES2-HS256+A128KW", 128 }, { "PBES2-HS384+A192KW", 192 }, { "PBES2-HS512+A256KW", 256 }
```
sberyozkin, line 23: These should be `A128KW` etc. `HS` in these PBES algos does not reflect the fact the password entered by the user may be different size... The scheme should be `AES`, please also see below for the proposal on having a map inside `KeyUtils` to avoid us having to parse the enum names, etc

sberyozkin:
Hey @ejba I've left quite a few suggestions to fix but IMHO you are really making it better and it is easier to work with your PR because it shows where we can improve in the code in general, thanks for that :-) we are nearly there :-)

sberyozkin:
Hi Emanuel, @ejba, this is perfect, sorry for keeping changing my mind  during the review and thank you for your patience :-).
I think we can now build from here and start dealing with other algorithms and key types as needed, but for now we have a good basis IMHO, thank you.
The only last bit which I'd like to ask for is to make those KeyUtils static maps package protected for now to give us some flexibility going forward...
Thanks

sberyozkin:
Please also rebase

sberyozkin (dismissed):
Looks good, waiting for one last minor update

sberyozkin (approved):

PR #403, opened 2021-01-29
Enforce types of the directly supported claims when they are set as Objects

### LINKED ISSUE #391: JwtClaimsBuilder fails to enforce the correct type of directly supported claims

sberyozkin:
Fixes #391

I've been thinking for a while what exactly would be the right course of action when someone types `.claim("sub", 1)` and decided that instead of silently converting to String and causing some potentially difficult to diagnose problems it would be better to enforce the correct types for only those claims which are well-known (ex JWT and Mp JWT specs) but more importantly those which are directly (have the dedicated setters) supported by this API - this will also ensure the consistent handling of such claims, whether they set in a typed way or as Objects.

A few simple claims must be Strings, `aud` or `groups` - Strings or Collections of Strings, `iat` and `exp` - `long` or `Instant`.

Also improved the documentation. And added `claim` setters accepting `Claims` as a claim name

### DIFF
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/Jwt.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/Jwt.java
@@ -1,5 +1,6 @@
 package io.smallrye.jwt.build;
 
+import java.time.Instant;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


radcortez (approved):

MikeEdgar (approved):

sberyozkin:
@radcortez @MikeEdgar thanks

PR #405, opened 2021-01-29
Fix configuration key location inconsistency and deprecate the inner-sign none signatures

### LINKED ISSUE #395: Fix configuration key location inconsistency
```
smallrye.jwt.encrypt.key-location
smallrye.jwt.sign.key-location
smallrye.jwt.verify.key.location
mp.jwt.verify.publickey.location
```

Configuration keys with dashes should be replaced with dots, making them all consistent and ensuring that smallrye is consistent with microprofile convention.

Probably I'm not the first person who has copied and modified the verify configuration and then spent some time debugging why the sign config appears to be unset.

Probably desirable to support old config keys for awhile to prevent breaking backwards compatibility.

### LINKED ISSUE #404: Deprecate inner-sign option where non-secured JWT is encrypted
I've been confident it has been already done - may be I've only set this task in the currently stale MP JWT PR.

So `.innerSign().encrypt(...)` supports an option where no signing keys have been configured in which a non-secured JWT is encrypted. This option is formally available in the JOSE space, but it is one option too many for this API, and it makes the point of inner-sign lost , and is not different to simply doing the direct encryption.

So I'm going to deprecate it and drop in 3.0.0

sberyozkin:
Fixes #395 
Fixes #404 

this PR aligns the sign/encrypt key properties with the convention used for configuring the verification/decryption keys, deprecates an inner-sign none signature mode which can only add to the confusion, and cleans up some docs

### DIFF
--- a/doc/modules/ROOT/pages/generate-jwt.adoc
+++ b/doc/modules/ROOT/pages/generate-jwt.adoc
@@ -140,8 +140,10 @@ Smallrye JWT supports the following properties which can be used to customize th
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
-|smallrye.jwt.encrypt.key-location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called.
-|smallrye.jwt.sign.key-location|none|Location of a key which will be used to sign the claims when either a no-argument sign() or innerSign() method is called.
+|smallrye.jwt.encrypt.key.location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called.
+|smallrye.jwt.encrypt.key-location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called. This property is deprecated and will be removed in the next major release.
+|smallrye.jwt.sign.key.location|none|Location of a key which will be used to sign the claims when either a no-argument sign() or innerSign() method is called.
+|smallrye.jwt.sign.key-location|none|Location of a key which will be used to sign the claims when either a no-argument sign() or innerSign() method is called. This property is deprecated and will be removed in the next major release.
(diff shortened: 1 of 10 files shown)


MikeEdgar (approved):

sberyozkin:
Hi Mike @MikeEdgar thanks; I'm going to rebase - it will drop the approval - but I'll keep it in mind :-)

doc/modules/ROOT/pages/generate-jwt.adoc
```diff
@@ -140,8 +140,8 @@ Smallrye JWT supports the following properties which can be used to customize th
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
-|smallrye.jwt.encrypt.key-location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called.
-|smallrye.jwt.sign.key-location|none|Location of a key which will be used to sign the claims when either a no-argument sign() or innerSign() method is called.
+|smallrye.jwt.encrypt.key.location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called.
```
radcortez, line 143: I'm wondering if we should keep the deprecated key here? But with a mark that is deprecated?

radcortez (approved):

doc/modules/ROOT/pages/generate-jwt.adoc
sberyozkin: @radcortez Hi Roberto, since it is not related to MP JWT at the moment, I thought of following a somewhat aggressive deprecation policy. So the old property stays in 2.4.x, but will be gone in 3.0.0 - otherwise we'd have to carry it around for a long time. That said, these are 2.4.3 docs which are being changed in this PR so yes, good point, let me add this note here as well (I've been thinking only about updating the JavaDocs)

sberyozkin: @radcortez I've updated the docs - noted that the 2 other properties are deprecated and will be removed - but I'll also 'lost' your approval with this update - please suggest some further doc updates if you'd prefer, otherwise I'll merge once it is green

PR #406, opened 2021-01-29
Reuse Jose4j HttpsJwk class for managing HTTP based JWK sets

### LINKED ISSUE #397: Behave identically for http jwks url
At the moment the code behaves different depending if jwks url is configured as http or as https.

In our case we are running the jwks provider in the same Kubernetes tenant as the backend. The provider is not able to provide the keyset over an https url. And because the traffic keeps inside the tenant it is also not needed for us from a security perspective.
But if I configure the keyset location with http url in smallrye-jwt, the refresh mechanism is not initialized and so JWT verification fails after the next key rotation.

Is there a specific reason for the different behavior? It would be great if https jwks would work identically to https?

sberyozkin:
Fixes #397

The case described at #397 is worth addressing IMHO.
This PR does it by re-using `HttpsJwks` - this sounds like a perfect anti-pattern :-), I rely here on the fact that `HttpsJwks` does not enforce the location is HTTPS based, it is more like a recommendation encoded into the class name as opposed to a strict requirement. But rather than copying and pasting `HttpsJwks` into our own copy of `HttpJwks` I'd rather have the single code dealing with the remote HTTP(S) sets - and only resort to copy if `Jose4j` starts enforcing HTTPS in `HttpsJwks` - which I really don't expect to happen any time soon as it will likely break some deployments.
(I might consider doing a Jose4J PR to have `HttpJwks` and `HttpJwks` extending the same shared code)   

Obviously we don't want the users to pull the sets from some random URLs but MP JWT itself requires we support HTTP JWK sets, so all this PR does it ensures HTTP based sets are treated the same way as HTTPS based sets.


MikeEdgar (approved):

sberyozkin:
Mike, have lost your approval again :-) due to the rebase, I'd just like Roberto to check too.

By the way I've been thinking about copying instead but I'm not sure it would be better :-), as any changes Jose4J may do to HttpsJwks may be missed so we'd have 2 diverging sources.... So I guess the current PR update is not a bad compromise :-)

MikeEdgar (approved):
I think using jose4j for both protocols is a safe bet. As you said, a change there would like break others as well.

radcortez (approved):

PR #413, opened 2021-02-04
Remove deprecated JWT Build API inner-sign none signature support and key location properties

sberyozkin:
This is the very last PR into the `mpjwt12` branch. It affects `implementation/jwt-build` module only.
Support for the deprecated `none` innger-sign algorithm is dropped, and the deprecated  sign/encrypt `key-location` properties. are also dropped.

### DIFF
--- a/doc/modules/ROOT/pages/generate-jwt.adoc
+++ b/doc/modules/ROOT/pages/generate-jwt.adoc
@@ -141,9 +141,7 @@ Smallrye JWT supports the following properties which can be used to customize th
 |===
 |Property Name|Default|Description
 |smallrye.jwt.encrypt.key.location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called.
-|smallrye.jwt.encrypt.key-location|none|Location of a key which will be used to encrypt the claims or inner JWT when a no-argument encrypt() method is called. This property is deprecated and will be removed in the next major release.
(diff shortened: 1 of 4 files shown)


sberyozkin:
Hi @radcortez @MikeEdgar @darranl Hi - I would like to have at least 1 approval if possible, so that I can get RC5 early next week. It is totally safe IMHO :-). For the build API itself it is a good clean up as well - inner-sign (none sig) + encrypt is just a more expensive way of doing the encryption only - and as such it is effectively a duplication.

radcortez (approved):

radcortez:
Done!

darranl:
@sberyozkin How long have these configuration options been available?  I am not sure if we exposed these in WildFly already.

sberyozkin:
@darranl Hi, For a while. But as I said these are not MP JWT 1.2 (or 1.1.1)  related properties. My understanding the vendor/library specific properties are not exposed in the current WildFly integration and more specifically, the API for generating the tokens which these properties related to is not even integrated AFAIK
thanks

darranl:
+1 I think this sounds like a safe deprecation to remove.

sberyozkin:
Unfortunately I've lost the history after moving this code into a standalone module - my fault. These properties are at least 6 months 'old'

sberyozkin:
@darranl thanks Darran, I'll take your `+1` as an approval :-).
@MikeEdgar, I'd like to try RC5 this evening so I'll go ahead with the merge
thanks all

PR #416, opened 2021-02-07
Complete the removal of smallrye.jwt.sign.key-location and smallrye.jwt.encrypt.key-location properties

sberyozkin:
@radcortez @darranl @MikeEdgar 

This PR follows #413 and actually completes the removal of these 2 deprecated properties in 3.0.0. I've run Quarkus `microprofile-4` branch and I did not see the failing tests which rely on those deprecated properties so I realized the previous PR was not complete. Roberto, Darran, I've got your approval to the earlier PR so I'll go ahead with merging this one once it passes the builds - the updates are very simple - the drop the properties typed as `DEPRECATED_...` in the code :-)


kiraka42:
Is it possible to set in smallrye.jwt.sign.key-location a String instead of a .PEM file ? like .mp.jwt.verify.publickey=String I did not found anything about it

PR #424, opened 2021-02-23
Merge the mpjwt12 branch to master

sberyozkin:


.github/workflows/build.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - mpjwt12
```
sberyozkin, line 7: I can follow up with updating it to `2.4.x` if needed

PR #425, opened 2021-02-23
Merge 'mpjwt12' branch to 'master' 

sberyozkin:


sberyozkin:
`project.xml` is not not changed, so the question is will github action try to release or not :-)

sberyozkin:
I guess even it does it will fail to release because `project.xml` points to `2.4.3` which has already been released

darranl:
@sberyozkin I am not sure where best to discuss this so hijacking your PR ;-)

We may shortly need to consider moving on to the next minor version bump as we may need to consider bug fixes to 3.0 without pulling in too many additional changes.

sberyozkin:
@darranl Hi Darran, np, I was going to request a review from you and others anyway once I get it green; i suggested this merge in our forum discussion, so thought there were no objections.
I think the only extra fix/change which will be picked up is this one:
https://github.com/smallrye/smallrye-jwt/pull/415

The only other changes which went into 2.4.x were the version updates which were dismissed during the conflict resolution

What do you think ? Would you like to discuss it in the forum ?

sberyozkin:
@darranl OK, let me continue the forum discussion as well

darranl:
@sberyozkin FYI for now I think I have enough time to pull in a 3.0.1 release if that is where you are heading, just after that there may be more scrutiny for the contents of the component upgrades.

sberyozkin:
@darranl No, I was only planning to make `mpjwt12` the master branch, no plans for an immediate 3.0.1 release on my end

darranl:
Ok thanks @sberyozkin - it is probably not directly related to this PR anyway - just some of the other PRs in the queue may be better in a minor bump instead of a micro.

sberyozkin:
Hi @darranl sure, I think going forward I will be sending emails to the forum just before attempting to do some next 2.4.x or 3.x release and we can all review the PRs, version numbers etc...
OK, sounds like you are OK with this PR then so far :-), let me request a review from yourself, Roberto and Mike

sberyozkin:
Quality check has failed - 3.0.1-SNAPSHOT contains an `==` string comparison - I did not see it but this is safe since effectively the pointers are compared there,  there is a local var which is assigned to one of the configured values... It can be replaced with `equals` now or later

radcortez:
> Quality check has failed - 3.0.1-SNAPSHOT [contains](https://github.com/smallrye/smallrye-jwt/blob/mpjwt12/implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java#L505) an `==` string comparison - I did not see it but this is safe since effectively the pointers are compared there, there is a local var which [is assigned](https://github.com/smallrye/smallrye-jwt/blob/mpjwt12/implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java#L409) to one of the configured values... It can be replaced with `equals` now or later

It is ok. You can add a comment in the source to ignore if you want.

Also, regarding the release if you open the merge from a fork it won't work. So another way to be on the safe side :)

MikeEdgar (approved):

sberyozkin:
@MikeEdgar, @radcortez OK, thanks; I'll close this PR again and re-create from my fork; I'll keep Mike's review in mind as well

PR #426, opened 2021-02-24
Merge the 'mpjwt12' branch to master

sberyozkin:


sberyozkin:
@radcortez Hey Roberto, opening a PR from the fork did not deter the `JWT` code owner from insisting on being one of the reviewers :-), I'll just trust it will be clever enough not to try to release given that after `project.xml` has remained unchanged after the conflict resolution. And if tries to release it will fail because `2.4.3` is already out

radcortez (approved):

radcortez:
> @radcortez Hey Roberto, opening a PR from the fork did not deter the `JWT` code owner from insisting on being one of the reviewers :-), I'll just trust it will be clever enough not to try to release given that after `project.xml` has remained unchanged after the conflict resolution. And if tries to release it will fail because `2.4.3` is already out

Yes, but in this case, forks don't have access to secrets, so if he tries to do a release it will fail :)

radcortez:
But I think it is fine. The job for pre-release didn't run:
https://github.com/smallrye/smallrye-jwt/actions/workflows/pre-release.yml

The last one was from the previous PR.

sberyozkin:
@radcortez oh great, that is definitely safer

sberyozkin:
@radcortez Ok, cleaned up `build.yml` as well

sberyozkin:
@radcortez @MikeEdgar Merging now since both of you approved, CC @darranl

sberyozkin:
@MikeEdgar @ejba's fix was also picked up since you approved, it is an isolated KeyUtils update

PR #429, opened 2021-02-25
Issue#427 DefaultJWTTokenParser#parseClaims Exception Handling

### LINKED ISSUE #427: DefaultJwtTokenParser#parseClaims Exception Handling
Hi all,

This issue relates to #372. We just upgraded WildFly to version 3.0.0 for smallrye-jwt, so I am finally able to test [WFLY-13164](https://issues.redhat.com/browse/WFLY-13164) with the fix I submitted. However, I realized we might have missed a small case in the exception handling for [DefaultJwtTokenParser#parseClaims](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java#L154).

When a invalid public key has been configured in the `mp.jwt.verify.publickey.location`, an `UnresolvableKeyException` should be reported. However, note that if this exception arises inside the call to the JOSE method [JWTConsumer#process](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java#L141), then the `UnresolvableKeyException` gets wrapped in a `InvalidJwtException` by the JOSE method. 

To see how the UnresolvableKeyException gets wrapped, I am leaving some links to the Jose code base: 
[0] [DefaultJwtTokenParser#parseClaims ](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java#L141) calls `jwtConsumer#process`
[1]  [JwtConsumer#process](https://bitbucket.org/b_c/jose4j/src/7ec0ef06c34e5f7ab70780548e6445e2ecf38107/src/main/java/org/jose4j/jwt/consumer/JwtConsumer.java#lines-316) gets called 
[2] [JwtConsumer#processContext](https://bitbucket.org/b_c/jose4j/src/7ec0ef06c34e5f7ab70780548e6445e2ecf38107/src/main/java/org/jose4j/jwt/consumer/JwtConsumer.java#lines-433) gets called 
[3] [JwtConsumer#processContext](https://bitbucket.org/b_c/jose4j/src/7ec0ef06c34e5f7ab70780548e6445e2ecf38107/src/main/java/org/jose4j/jwt/consumer/JwtConsumer.java#lines-213) throws `UnresolvableKeyException` in `verificationKeyResolver#resolveKey`
[4] Jose wraps the exception in [InvalidJwtException](https://bitbucket.org/b_c/jose4j/src/7ec0ef06c34e5f7ab70780548e6445e2ecf38107/src/main/java/org/jose4j/jwt/consumer/JwtConsumer.java#lines-271)

Therefore, once we reach the exception handling in [DefaultJwtTokenParser#parseClaims,](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java#L154) it logs that the token is invalid (whereas it should log that the public key is unresolvable) and then that JOSE exception gets passed wrapped as a `ParseException` to [JWTHTTPAuthenticationMechanism#validateRequest](https://github.com/smallrye/smallrye-jwt/blob/master/implementation/jwt-http-mechanism/src/main/java/io/smallrye/jwt/auth/mechanism/JWTHttpAuthenticationMechanism.java#L84) where the fix I added won't report an error 500 as `e.getCause()` is an `InvalidJwtException` whereas `e.getCause().getCause()` would be an `UnresolvableKeyException`, which is what we are looking for. 

I was wondering whether it would be appropriate to send an `UnresolvableKeyException` in the `DefaultJWTTokenParser` instead in this specific case?

Thanks so much for your help!

SoniaZaldana:
Fixes #427 

When a invalid public key has been configured in the `mp.jwt.verify.publickey.location`, an `UnresolvableKeyException` should be reported. However, if this exception arises inside the call to the JOSE method, then the `UnresolvableKeyException` gets wrapped in a `InvalidJwtException` by the JOSE method, resulting in the wrong error code being sent in `JWTHTTPAuthenticationMechanism.`

### DIFF
--- a/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
+++ b/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
@@ -152,8 +152,13 @@ public class DefaultJWTTokenParser {
 
             return jwtContext;
         } catch (InvalidJwtException e) {
-            PrincipalLogging.log.tokenInvalid();
-            throw PrincipalMessages.msg.failedToVerifyToken(e);
+            if (e.getCause() instanceof UnresolvableKeyException) {
+                PrincipalLogging.log.verificationKeyUnresolvable();
+                throw PrincipalMessages.msg.failedToVerifyToken(e.getCause());
+            } else {
+                PrincipalLogging.log.tokenInvalid();
+                throw PrincipalMessages.msg.failedToVerifyToken(e);
+            }
         } catch (UnresolvableKeyException e) {
             PrincipalLogging.log.verificationKeyUnresolvable();
             throw PrincipalMessages.msg.failedToVerifyToken(e);
--- a/testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/DefaultJWTParserTest.java
+++ b/testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/DefaultJWTParserTest.java
@@ -11,7 +11,7 @@ import javax.crypto.SecretKey;
 import org.eclipse.microprofile.jwt.JsonWebToken;
 import org.jose4j.jwk.JsonWebKey;
 import org.jose4j.jwk.PublicJsonWebKey;
-import org.jose4j.jwt.consumer.InvalidJwtException;
+import org.jose4j.lang.UnresolvableKeyException;
 import org.junit.Test;
 
 import io.smallrye.jwt.algorithm.KeyEncryptionAlgorithm;
@@ -75,9 +75,9 @@ public class DefaultJWTParserTest {
         JWTAuthContextInfo config = new JWTAuthContextInfo("/certificate.pem", "https://server.example.com");
         config.setVerifyCertificateThumbprint(true);
         JWTParser parser = new DefaultJWTParser(config);
-        ParseException thrown = assertThrows("InvalidJwtException is expected",
+        ParseException thrown = assertThrows("UnresolvableKeyException is expected",
                 ParseException.class, () -> parser.parse(jwtString));
-        assertTrue(thrown.getCause() instanceof InvalidJwtException);
+        assertTrue(thrown.getCause() instanceof UnresolvableKeyException);
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


sberyozkin:
@SoniaZaldana Thanks; FYI, `testsuite/basic` fails, please update.

I think it is a correct decision - this is the case when JWT may well be valid, the key is unresolvable, so indeed it is correct to throw this exception unwrapped - cause of `InvalidJwtException` should be a concrete JWT problem which caused it. Technically it is a breaking change, but putting it into a different surface, I'd qualify it as a bug fix, hence the backward compatibility concerns do not quite apply.
We can do `3.1.0` though, @darranl is is what you meant when you were referring to a possible `3.1.0` ?

SoniaZaldana:
hi @sberyozkin I was taking a closer look at one of the test failures `DefaultJwtParserTest#testParseWithConfiguredCertAndThumbprintMissing`. In this test case, with the code that I added, it is reporting an UnresolvableKeyException when the `JWS `is missing the `X5T` or `X5T#S256` header. I am wondering whether this would be considered a concrete JWT problem where an `InvalidJWTException` might be more appropriate due to the missing headers or whether we still consider this a potentially valid JWT where we were just unable to resolve the public key?

sberyozkin:
@SoniaZaldana Sorry I have missed this comment, I thought I checked it during the recharge day :-)

No, this is still an UnresolvedKeyException as that test is about a case where a thumbprint is acting as a `kid` for the X509 resolver to match it against the pre-calculated thumbprint from the `x5c`chain JWK property - the key would then come from the matched `x5c`. It is not a standard verification process as well - it currently depends on the smallrye-jwt specific property.

SoniaZaldana:
Thanks @sberyozkin! I updated the PR and all tests pass now.

sberyozkin:
@SoniaZaldana Nice work, thanks for your careful analysis of the failing tests as well

sberyozkin (approved):

PR #431, opened 2021-02-26
Add smallrye.jwt.new-token.override-matching-claims property

### LINKED ISSUE #420: Make the order in which JWT build configured claim values are applied configurable
For example we have `smallrye.jwt.new-token.audience` - if this property is set then its value will be picked up if no `aud` is set.

But take this example where we want to resign but also set a new audience (and also new issuer): 
```
smallrye.jwt.new-token.audience=downstream-service-audience
smallrye.jwt.new-token.issuer=this-endpoint

@Inject JsonWebToken jwt;
// resign the current jwt token with a new key and propagate
String jwt = Jwt.claims(jwt).sign();
``` 

Unfortunately the new audience and issuer won't be picked if the audience and the issuer are already set - issuer for ex is nearly always set in JWT tokens. The only option is to reset them manually in the code.

So for a case like this one, to enable a super-fast re-signing and resetting of the key properties via the configuration, the order (priority) of the configured claims properties should be optionally configurable. Something like this:

```
smallrye.jwt.new-token-properties-preferred=true
smallrye.jwt.new-token.audience=downstream-service-audience
smallrye.jwt.new-token.issuer=this-endpoint

@Inject JsonWebToken jwt;
// resign the current jwt token with a new key and propagate
String jwt = Jwt.claims(jwt).sign();

// Bingo, this token has got the new issuer and audience properties
```

sberyozkin:
Fixes #420

This PR supports making the `new-token.issuer` and `new-token.audience` override the existing claims, such that, for example, in `quarkus-oidc-token-propagation` I can code:
```
@Inject JsonWebToken jwt;

String newJwt = Jwt.claims(jwt).sign();
```

and doc that the users - if they need this feature, should only configure the new signing key, and/or issuer and audience
etc.

Totally isolated to `jwt-build`

### DIFF
--- a/doc/modules/ROOT/pages/generate-jwt.adoc
+++ b/doc/modules/ROOT/pages/generate-jwt.adoc
@@ -145,4 +145,5 @@ Smallrye JWT supports the following properties which can be used to customize th
 |smallrye.jwt.new-token.lifespan|300|Token lifespan in seconds which will be used to calculate an `exp` (expiry) claim value if this claim has not already been set.
 |smallrye.jwt.new-token.issuer|none|Token issuer which can be used to set an `iss` (issuer) claim value if this claim has not already been set.
 |smallrye.jwt.new-token.audience|none|Token audience which can be used to set an `aud` (audience) claim value if this claim has not already been set.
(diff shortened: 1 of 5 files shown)


MikeEdgar (approved):
Looks good!

sberyozkin:
@MikeEdgar thanks, I think a single approval is enough for this update, CC @radcortez Roberto - I'll CC you in my Quarkus PR :-)

sberyozkin:
Just updated the branch

PR #433, opened 2021-02-26
Remove deprecated KeyUtils and ResourceUtils

sberyozkin:
Unfortunately these are another 2 deprecated classes I forgot to remove before releasing `3.0.0`.
Both classes are in `jwt-common` now, in a new package `io.smallrye.jwt.util`, the removed classes only delegate to the new classes and have no any specific implementation of their own.

I'd not be too concerned if we decided to keep them, I feel it is still safe to remove them instead of carrying this code, but if someone is concerned them let me know please and I'll keep them


MikeEdgar (approved):

sberyozkin:
@MikeEdgar thanks, I'll wait for a few days in case @darranl and @SoniaZaldana might have some concerns - I'd not expecf this deprecated code be used anywhere in WildFly but just in case :-)

sberyozkin:
@darranl @SoniaZaldana merging now - any concerns - let me know please here or on the forum

PR #438, opened 2021-03-18
Documentation: Replace deprecated configuration keys

enqueue:
`smallrye.jwt.encrypt.key-location` and `smallrye.jwt.sign.key-location` are deprecated, so the documentation should refer to the new configuration keys.

Related #413 #416


sberyozkin:
@enqueue Thanks for spotting it, I updated the table where the properties are referenced but forgot the comments :-)

sberyozkin (approved):

PR #441, opened 2021-03-25
Update GitHub workflows to use 'main'

sberyozkin:


sberyozkin:
@radcortez, Hi Roberto, it is not going to start a release ? It should be `3.1.0`, and `project.yml` still has the old values. Should I update it as well in one go ?

radcortez (approved):

radcortez:
The release is only triggered with a change to `project.yml` coming from a PR. If you want to update it and not do a release, just push the commit directly to the branch :)

sberyozkin:
@radcortez Oh yeah, good, this is the main branch so I can't push directly - but indeed I'll do it later this evening with a PR :-)

PR #454, opened 2021-04-27
Verify fully formatted PEM key can be accepted and invalid key causes UnresolvedKeyException

sberyozkin:
It is already TCK tested - just adding one extra local test to verify it


MikeEdgar (approved):
Looks good to me

PR #455, opened 2021-04-28
Use JWK 'alg' and 'kid' when creating JWT if these headers are not set

### LINKED ISSUE #321: Use JWK `kid` when generating the tokens if no `kid` is already set

sberyozkin:
Fixes #321.

This PR makes sure that `alg` and `kid` properties do not have to be set manually when `JWK` keys are used to sign or encrypt the claims and those keys have already either `alg` and/or `kid` set. 

`KeyUtils` have been updated a bit - none of that code is used by the server side MP JWT code.
Also CC @SoniaZaldana


MikeEdgar (approved):
Looks good

radcortez (approved):

sberyozkin:
@MikeEdgar @radcortez thanks

PR #466, opened 2021-05-20
Ensure JWT ID is unique when the claims or signature builder is reused

### LINKED ISSUE #463: Make sure JWT ID (jti) claim value is unique when the builder is reused
`jti` claim is set automatically if it has not been set - but if the builder is re-used then it will remain the same when the next token is created - which can cause a token rejection.
The last `jti` value needs to be saved - the builder is not expected to be thread safe and it is already doc-ed - it keeps the claims/headers state already

sberyozkin:
Fixes #463.

Note I've added a note that the encryption builder should not be reused - this is because it operated on a `String`, example, it can already be a signed JWT.
It does not mean one can not get a unique `jti` when encrypting the tokens, it is only about recommending to avoid (not very likely to happen) pattern:
```
JwtEncryptionBuilder builder = Jwt.upn("alice").innerSign();
builder.keyId("1").encrypt();
builder.keyId("2").encrypt();
```
the same can be done by simply doing 

```
Jwt.upn("alice").innerSign().keyId("1").encrypt();
Jwt.upn("alice").innerSign().keyId("2").encrypt();
```

or

```
JwtClaimsBuilder builder = Jwt.upn("alice");
builder.innerSign().keyId("1").encrypt();
builder.innerSign().keyId("2").encrypt();
```

### DIFF
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtClaimsBuilderImpl.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtClaimsBuilderImpl.java
@@ -266,7 +266,11 @@ class JwtClaimsBuilderImpl extends JwtSignatureImpl implements JwtClaimsBuilder,
     @Override
     public JwtEncryptionBuilder jwe() {
         JwtBuildUtils.setDefaultJwtClaims(claims, tokenLifespan);
-        return new JwtEncryptionImpl(claims.toJson());
+        try {
+            return new JwtEncryptionImpl(claims.toJson());
+        } finally {
+            removeJti();
+        }
     }
 
     @SuppressWarnings({ "rawtypes", "unchecked" })
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/JwtClaimsBuilder.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/JwtClaimsBuilder.java
@@ -17,6 +17,7 @@ import org.eclipse.microprofile.jwt.Claims;
  * <p>
  * JwtClaimsBuilder implementations must set the 'iat' (issued at time), 'exp' (expiration time)
  * and 'jti' (unique token identifier) claims unless they have already been set.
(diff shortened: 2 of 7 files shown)


implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtClaimsBuilderImpl.java
```diff
@@ -265,7 +265,7 @@ public JwtEncryptionBuilder innerSign() throws JwtSignatureException {
      */
     @Override
     public JwtEncryptionBuilder jwe() {
-        JwtBuildUtils.setDefaultJwtClaims(claims, tokenLifespan);
+        lastJti = JwtBuildUtils.setDefaultJwtClaims(claims, lastJti, tokenLifespan);
         return new JwtEncryptionImpl(claims.toJson());
     }
```
MikeEdgar, line 270: Would it make sense to instead remove the `jti` claim from the `JwtClaims` before returning rather than keeping track of the previous value? Just mentioning in case it hasn't been considered.

MikeEdgar (commented):
One question/suggestion, @sberyozkin

implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtClaimsBuilderImpl.java
sberyozkin: Hi Mike @MikeEdgar It might've crossed my mind but I did not look into it. May be it would be a cleaner solution. It might not be possible to ensure that in all the cases the uniqueness is assured, so perhaps just removing it would do; I'll have a look.

sberyozkin: @MikeEdgar I've fixed it to apply your idea, which is simpler and easier to understand, the original approach I took was looking not good at all to me this morning when I returned to changing it :-), thanks

MikeEdgar (approved):
Thanks @sberyozkin. Looks good to me!

PR #467, opened 2021-05-25
Cache configured PEM keys and check configured kid for JWK keys

### LINKED ISSUE #462: Optionally cache the configured keys used to sign/encrypt the claims
When we have `Jwt.claim("a", "b").sign()` or `Jwt.claim("a", "b").encrypt()` the configured keys are loaded from the file system/etc.
If the same builder is used to generate multiple tokens then it is suboptimal to load it every time

sberyozkin:
Fixes #462

After looking at the code again I thought of addressing #462 by only caching PEM configured keys since there could be no ambiguities but decided to avoid caching JWK keys as their key identifiers can vary between the requests, in theory at least, so the only optimization when dealing with the JWK keys is when they are loaded from a multi-key JWK set - where new `smallrye.jwt.sign.key.id`/`smallrye.jwt.encrypt.key.id` can help with simplifying the code - as setting `kid` is a technical operation

I'm planning to release `3.2.0` after this PR is merged

### DIFF
--- a/implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtBuildConfigSource.java
+++ b/implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtBuildConfigSource.java
@@ -8,54 +8,67 @@ import java.util.Set;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
 
-public class JwtBuildConfigSource implements ConfigSource {
-    private static final String SIGN_KEY_LOCATION_PROPERTY = "smallrye.jwt.sign.key.location";
-    private static final String ENC_KEY_LOCATION_PROPERTY = "smallrye.jwt.encrypt.key.location";
+import io.smallrye.jwt.build.impl.JwtBuildUtils;
 
-    boolean signingKeyAvailable = true;
+public class JwtBuildConfigSource implements ConfigSource {
+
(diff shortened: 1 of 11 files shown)


implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtBuildConfigSource.java
```diff
@@ -80,8 +93,10 @@ public void setAudiencePropertyRequired(boolean audiencePropertyRequired) {
 
     @Override
     public Set<String> getPropertyNames() {
-        return new HashSet<>(Arrays.asList("smallrye.jwt.sign.key-location",
-                "smallrye.jwt.encrypt.key-location",
+        return new HashSet<>(Arrays.asList("smallrye.jwt.sign.key.location",
```
radcortez, line 96: We can probably static init this (or just init it in the constructor)

radcortez (approved):

implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtBuildConfigSource.java
sberyozkin: Hi Roberto @radcortez I did some cleanup of this test config source (with a positive side-effect is that I've got all supported properties now declared in a single place in `JwtBuildUtils`).
I've just stopped short of refactoring `getValue` for it not be implemented in terms of `getProperties` as it is mostly conditional there per different test requirements so it would be also quite messy in `getValue()` :-)

radcortez (approved):

PR #475, opened 2021-06-11
Accept tokens without iat if time-to-live is 0 or negative

### LINKED ISSUE #469: Let users disable the issued at claim requirement
A Quarkus user has reported that with smallrye-jwt `3.x` the tokens without `iat` are rejected.

MP JWT 1.2 enforces an MP JWT 1.1.1 requirement that `iat` is one of the required claims (not all of the claims from that MP JWT 1.1.1 list are enforced) but `iat` is one of the very essential claims so it is enforced.

The problem is the user is working with a 3rd party issuer which can not be changed in a short term.

So we need to add a smallrye-jwt specific property to let the users make it optional

CC @@xwoker

sberyozkin:
Fixes #469 

I've ended up reusing the token from [this JWS spec example](https://datatracker.ietf.org/doc/html/rfc7515#appendix-A.1) for the tests, since both MP JWT TCK utiliity code and JWT Build insist on setting `iat` and I'm not keen at the moment to introduce a new JWT Build property to let it avoid setting `iat`.
The token in that example is already expired and has no `sub` or other name related claims so I've updated the test to use the `iss` claim as a `sub` one and added a massive expiry grace time :-)


MikeEdgar (approved):

sberyozkin:
thanks @MikeEdgar

PR #476, opened 2021-06-11
Support the conversion of JsonObject to custom types

sberyozkin:
Hi Roberto, @radcortez, this is another draft PR, focusing for now on `JsonWebToken.getClaim(String name)` and the conversion of complex claims to start with, it can be aligned somehow with your PR where an individual claim injection is supported.
 
The immediate issue is that it is only `Converter<T>.convert(String)` at the moment, can you please update it to have 2 type variables for the input to `convert` be customized as needed ?

thanks


radcortez:
Sure, let me have a look.

radcortez:
@sberyozkin started some work on this. The API looks like this:
https://github.com/smallrye/smallrye-converters/pull/23/files#diff-5e81a214b8baf008f2e6d28449a412b085965297d6561ad5042b16a76b048c4dR28-R51

What do you think?

sberyozkin:
@radcortez Thanks Roberto, I'll try to experiment in the next few days

PR #494, opened 2021-08-20
Using the correct default key encryption algorithm

### LINKED ISSUE #493: JwtEncryptionImpl uses the wrong default key encryption algorithm
It is `RSA-OAEP` now which is also doc-ed in `JwtEncryption` but the implementation continues setting the default algorithm (if none is already set) to `RSA-OAEP-256` - this causes the confusion since on the server code does default to `RSA-OAEP`

sberyozkin:
Fixes #493

Right now the docs say it is `RSA_OAEP` while the code continues using `RSA_OAEP_256` by default - and it causes confusions when MP JWT server, assuming `RSA_OAEP` is indeed a default one, fails to decrypt.

I might do `3.3.0` after it as opposed to `3.2.2`.


MikeEdgar (approved):
Looks good to me.

sberyozkin:
Thanks @MikeEdgar

PR #515, opened 2021-10-29
Undeprecate smallrye.jwt.verify.algorithm for setting HMAC algorithms

### LINKED ISSUE #507: Undeprecate `smallrye.jwt.verify.key.algorithm` for customizing symmetric algorithms
`smallrye.jwt.verify.key.algorithm` is deprecated given that a standard `mp.jwt.verify.publickey.algorithm` is now available - with MP JWT focusing on RS or ES signatures. So if a user need to verify a token signed by a secret key then using `mp.jwt.verify.publickey.algorithm` to specify is `HS*` is not great.

The proposal is to un-deprecate `smallrye.jwt.verify.key.algorithm` and only warn it is deprecated if a user tries to use it to customize `RS*` or `ES*` algorithms

sberyozkin:
Fixes #507.

Quarkus user had to set `mp.jwt.verify.publickey.algorithm=HS512` which does not look correct even though it works, so I've un-deprecated the usage of `smallrye.jwt.verify.algorithm` for setting `HS256/384/512`

### DIFF
--- a/implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/ConfigMessages.java
+++ b/implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/ConfigMessages.java
@@ -11,8 +11,8 @@ import org.jboss.logging.annotations.MessageBundle;
 interface ConfigMessages {
     ConfigMessages msg = Messages.getBundle(ConfigMessages.class);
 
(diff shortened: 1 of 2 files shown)


MikeEdgar (approved):
Looks good to me

PR #526, opened 2021-11-22
Issue#253 Convert project to use Galleon based provisioning instead of a zip download and extraction.

darranl:
Additionally switch to WildFly 25.0.1.Final.

This is a follow up to https://github.com/smallrye/smallrye-jwt/pull/256

I have simplified this to the point that the provisioned server is very similar to the servlet distribution previously used.

The war which is deployed for the TCK is quite large, a follow up approach would then be to start to add Galleon layers so dependencies can move from the war ideally slimming the war to the point it only contains SmallRye artefacts. early contenders would be:

- CDI
- Rest

The EE Security integration could also be utilised which is closer to how WildFly pulls this in to enable JWT.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,9 @@
     <version.jose4j>0.7.9</version.jose4j>
     <version.resteasy>4.7.3.Final</version.resteasy>
     <version.mokito>4.1.0</version.mokito>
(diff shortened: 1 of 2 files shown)


sberyozkin (approved):
Nice work, thanks

PR #550, opened 2022-01-19
JWT Parser fixes

### LINKED ISSUE #538: In DefaultJWTTokenParser first keyResover wins no matter what authInfoContext specifies
I try to call an oidc secured service out of an oidc secured service. So the default `publicKeyLocation` is set to my IDP in order to verify tokens. When trying to validate the JWT from external service I cannot specify another `publicKeyLocation` in a dedicated `JWTAuthContextInfo`.

In https://github.com/smallrye/smallrye-jwt/blob/main/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java#L239 the `VerificationKeyResolver` should be returned. 

Now when using more than one IDP I create a jwtParser and pass an authContext to its parse method specifying another AuthInfo object with another publicKeyLocation but the method does not check if the existing `VerificationKeyResolver` matches the `authContextInfo` so I get the wrong keyResolver returned and `SRJWT07005: Failed to load a key from ...` error.

Is this a bug or am I doing something wrong?

### LINKED ISSUE #540: All configurations is loss when `JWTAuthContextInfo` bean is copied from an original proxy bean.
**Issue**
In copy constructor of `JWTAuthContextInfo`, direct access to private is used instead of accessor.

[JWTAuthContextInfo.java#L91L124](https://github.com/smallrye/smallrye-jwt/blob/main/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/JWTAuthContextInfo.java#L91L124)
```java
this.publicVerificationKey = orig.publicVerificationKey; // null; actual value is at orig.getPublicVeirificationKey()
...
```

In many solutions such as Quarkus DI, proxy bean is created where underling properties do not hold any values. Actual values are available via accessor. Hence this copy constructor would return an all null object.

**Suggestion**
Change the copy constructor to use accessors.

sberyozkin:
Fixes #540 (when `DefaultParser` is asked to parse with custom keys the copy construction of the context is done and the configured properties are lost - I have a Quarkus test verifying it)
Fixes #538 (when `DefaultParser` is asked to parse with the new key location on every request the cached `KeyLocationResolver` is used - so in this case a new factory instance has to be created - all works fine if new keys are loaded and passed directly as no key resolution is necessary in such cases)


sberyozkin:
I'll have a look, last minute changes

radcortez (approved):

sberyozkin:
Thanks @radcortez for both approvals

PR #551, opened 2022-01-19
Fix the problem with the zero expiry grace period

### LINKED ISSUE #546: Setting the expiry grace period to 0 fails to verify the expired tokens
A user has reported that when changing a grace period from the default 60 secs to 0 the expired tokens can be accepted - it will need to be reproduced with a test and fixed

sberyozkin:
Fixes #546


radcortez (dismissed):

sberyozkin:
@radcortez I'm going to rebase this PR - and then will merge

PR #556, opened 2022-01-26
Run TCK with Jetty

radcortez:
This allows running the TCK with Jetty

The motivation is to be able to run the TCK with Jakarta, since there is no release with source-based Jakarta for Wildfly.

https://github.com/smallrye/smallrye-jwt/tree/jakarta


sberyozkin (approved):
Hey @radcortez Missed it, thanks for looking into it

PR #574, opened 2022-03-14
Distinguish between unresolvable and invalid keys

### LINKED ISSUE #437: Distinguish between unresolvable and invalid keys
If we have a JWK key set, for example, coming from Keycloak and a token has a `kid` which does not match any JWK in this JWK key set then an `UnresolvableKeyException` will be thrown and - `500` reported in `jwt-http-mechanism`.
But this is a case of `401` - keys are available and valid - but the token is effectively invalid.
Some thinking is required how to do it cleanly

CC @darranl @SoniaZaldana

Skyllarr:
Fixes #437

### DIFF
--- a/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
+++ b/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
@@ -16,6 +16,7 @@
  */
 package io.smallrye.jwt.auth.principal;
 
+import java.io.IOException;
 import java.security.Key;
 import java.security.PublicKey;
 import java.security.cert.X509Certificate;
@@ -26,6 +27,7 @@ import org.jose4j.jwk.PublicJsonWebKey;
 import org.jose4j.jws.JsonWebSignature;
 import org.jose4j.jwx.JsonWebStructure;
 import org.jose4j.keys.resolvers.VerificationKeyResolver;
+import org.jose4j.lang.JoseException;
 import org.jose4j.lang.UnresolvableKeyException;
 
 import io.smallrye.jwt.KeyFormat;
@@ -63,6 +65,14 @@ public class KeyLocationResolver extends AbstractKeyLocationResolver implements 
         Key theKey = tryAsVerificationJwk(jws);
 
         if (theKey == null) {
+            try {
+                if (httpsJwks != null && httpsJwks.getJsonWebKeys() != null && jws != null
+                        && jws.getKeyIdHeaderValue() != null) {
+                    throw PrincipalMessages.msg.unmatchedTokenKidException();
+                }
+            } catch (JoseException | IOException e) {
+                // ignore, if JWK is unavailable this was logged previously
+            }
             reportUnresolvableKeyException(authContextInfo.getPublicKeyContent(), authContextInfo.getPublicKeyLocation());
         }
         return theKey;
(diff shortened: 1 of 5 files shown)


implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/JWTParser.java
```diff
@@ -38,7 +39,7 @@
      * @return JsonWebToken
      * @throws ParseException parse exception
      */
-    public JsonWebToken parse(final String token) throws ParseException;
+    public JsonWebToken parse(final String token) throws ParseException, UnresolvableKeyException;
```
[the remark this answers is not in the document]
sberyozkin: Hi @Skyllarr Default parser factory uses `ParseException` as a wrapper around the actual exception cause; so for now please don't add it, we can review it later, thanks

implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java
```diff
@@ -63,7 +63,7 @@ public Key resolveKey(JsonWebSignature jws, List<JsonWebStructure> nestingContex
         Key theKey = tryAsVerificationJwk(jws);
 
         if (theKey == null) {
-            reportUnresolvableKeyException(authContextInfo.getPublicKeyContent(), authContextInfo.getPublicKeyLocation());
+            throw PrincipalMessages.msg.invalidTokenKid();
```
Skyllarr, line 66: @sberyozkin Thanks for feedback! I updated my PR. I changed the exception here. It seems to me that if the reason why `(theKey == null)` is different than invalid kid, some other exception should have been thrown before this? 

Eg. when loading of keys was not successful because of a connection problem, the exception gets thrown [in the constructor](https://github.com/smallrye/smallrye-jwt/blob/main/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/KeyLocationResolver.java#L47) already. But let me know if this is wrong assumption.

sberyozkin: @Skyllarr Sorry for a delay - I wonder, would it be better to introduce an exception like `UnmatchedTokenKidException` and report it as a cause of `UnresolvableKeyException` only if it is the case when 1) the token has `kid` and it is JWK set or JWK and no match is found even after a refresh... Then instead of checking the message content, we just check for the cause...

Skyllarr: @sberyozkin Hi, I am sorry for my delay. I have introduced an `UnmatchedTokenKidException` exception that gets thrown when the JWKs are available and kid exists but the key was not found:

`if (httpsJwks != null && httpsJwks.getJsonWebKeys() != null && jws != null && jws.getKeyIdHeaderValue() != null)`

Let me know if that looks good. Thank you!

sberyozkin: @Skyllarr Hi, np at all, I've badly missed your update. I'll try to prioritize on getting it in asap, will take a bit more time but I'll have a look.

sberyozkin (approved):
Thanks @Skyllarr, reporting 401 (unverified token due to a non-matched kid) is indeed better IMHO, thanks for completing this PR

PR #582, opened 2022-04-13
Support for TLS certificates and hostname verification

### LINKED ISSUE #571: `mp.jwt.verify.publickey.location` needs to support TLS configuration
This URL's need to be able to configure TLS parameters (e.g. CA, Cert, Key).

We use an internal service (via a Vault plugin) that generates internal JWTs for micro-service authorization and provides public keys for verification via a standard JWKs endpoint. With TLS required for Vault we cannot load our verifications keys from this endpoint since we use a cluster local self-signed CA.

### LINKED ISSUE #581: jwt.verify.publickey.location property disable ssl hostname verify?
**i set jwt.verify.publickey.location property in my properties:**

mp.jwt.verify.publickey.location=https://userservice/.well-known/jwks.json
mp.jwt.verify.issuer=TelCloud

**and i'm use quarkus-smallrye-jwt, and it's version is 2.4.1.Final**

**but I've had such a mistake:**

Caused by: javax.net.ssl.SSLHandshakeException: No subject alternative DNS name matching userservice found.
        at java.base/sun.security.ssl.Alert.createSSLException(Unknown Source)
        at java.base/sun.security.ssl.TransportContext.fatal(Unknown Source)
        at java.base/sun.security.ssl.TransportContext.fatal(Unknown Source)
        at java.base/sun.security.ssl.TransportContext.fatal(Unknown Source)
        at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.checkServerCerts(Unknown Source)
        at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.onCertificate(Unknown Source)
        at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.consume(Unknown Source)
        at java.base/sun.security.ssl.SSLHandshake.consume(Unknown Source)
        at java.base/sun.security.ssl.HandshakeContext.dispatch(Unknown Source)
        at java.base/sun.security.ssl.HandshakeContext.dispatch(Unknown Source)
        at java.base/sun.security.ssl.TransportContext.dispatch(Unknown Source)
        at java.base/sun.security.ssl.SSLTransport.decode(Unknown Source)
        at java.base/sun.security.ssl.SSLSocketImpl.decode(Unknown Source)
        at java.base/sun.security.ssl.SSLSocketImpl.readHandshakeRecord(Unknown Source)
        at java.base/sun.security.ssl.SSLSocketImpl.startHandshake(Unknown Source)
        at java.base/sun.net.www.protocol.https.HttpsClient.afterConnect(Unknown Source)
        at java.base/sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(Unknown Source)
        at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(Unknown Source)
        at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream(Unknown Source)
        at java.base/java.net.HttpURLConnection.getResponseCode(Unknown Source)
        at java.base/sun.net.www.protocol.https.HttpsURLConnectionImpl.getResponseCode(Unknown Source)
        at org.jose4j.http.Get.get(Get.java:80)
        at org.jose4j.jwk.HttpsJwks.refresh(HttpsJwks.java:204)
        at io.smallrye.jwt.auth.principal.AbstractKeyLocationResolver.isHttpsJwksInitialized(AbstractKeyLocationResolver.java:100)
        at io.smallrye.jwt.auth.principal.KeyLocationResolver.initializeKeyContent(KeyLocationResolver.java:89)
        at io.smallrye.jwt.auth.principal.KeyLocationResolver.<init>(KeyLocationResolver.java:45)
        ... 81 more
Caused by: java.security.cert.CertificateException: No subject alternative DNS name matching userservice found.
        at java.base/sun.security.util.HostnameChecker.matchDNS(Unknown Source)
        at java.base/sun.security.util.HostnameChecker.match(Unknown Source)
        at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(Unknown Source)
        at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(Unknown Source)
        at java.base/sun.security.ssl.X509TrustManagerImpl.checkTrusted(Unknown Source)
        at java.base/sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(Unknown Source)
        ... 103 more

**so i want to diasble ssl hostname verify, who can help me?**

sberyozkin:
Fixes #571
Fixes #581

Currently only certificates in PEM format are supported and read from the file, but in the future we can support loading them from the truststores 

CC @kdubb

### DIFF
--- a/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/AbstractKeyLocationResolver.java
+++ b/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/AbstractKeyLocationResolver.java
@@ -22,7 +22,12 @@ import java.security.Key;
 import java.security.cert.X509Certificate;
 import java.util.Base64;
 import java.util.List;
+import java.util.Set;
 
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLSession;
+
+import org.jose4j.http.Get;
 import org.jose4j.jwk.HttpsJwks;
 import org.jose4j.jwk.JsonWebKey;
 import org.jose4j.jwk.OctetSequenceJsonWebKey;
@@ -39,6 +44,7 @@ import io.smallrye.jwt.util.ResourceUtils.UrlStreamResolver;
  * This implements the MP-JWT 1.1 mp.jwt.verify.publickey.location config property resolution logic
  */
 public class AbstractKeyLocationResolver {
+
     private static final String HTTP_SCHEME = "http:";
     private static final String HTTPS_SCHEME = "https:";
 
@@ -85,14 +91,37 @@ public class AbstractKeyLocationResolver {
         return jws.getHeaders().getStringHeaderValue(JsonWebKey.KEY_ID_PARAMETER);
     }
 
-    protected HttpsJwks initializeHttpsJwks(String location) throws IOException {
+    protected HttpsJwks initializeHttpsJwks(String location)
+            throws IOException {
         PrincipalLogging.log.tryCreateKeyFromHttpsJWKS();
+        HttpsJwks theHttpsJwks = getHttpsJwks(location);
+        if (location.startsWith(HTTPS_SCHEME)) {
+            Get httpGet = getHttpGet();
+            if (authContextInfo.isTlsTrustAll()) {
+                httpGet.setHostnameVerifier(new TrustAllHostnameVerifier());
+            } else if (authContextInfo.getTlsTrustedHosts() != null) {
+                httpGet.setHostnameVerifier(new TrustedHostsHostnameVerifier(authContextInfo.getTlsTrustedHosts()));
+            }
+            if (authContextInfo.getTlsCertificatePath() != null) {
+                httpGet.setTrustedCertificates(loadPEMCertificate(readKeyContent(authContextInfo.getTlsCertificatePath())));
+            }
+            theHttpsJwks.setSimpleHttpGet(httpGet);
+        }
+        return theHttpsJwks;
+    }
+
+    protected HttpsJwks getHttpsJwks(String location) {
         HttpsJwks theHttpsJwks = new HttpsJwks(location);
         theHttpsJwks.setDefaultCacheDuration(authContextInfo.getJwksRefreshInterval().longValue() * 60L);
         return theHttpsJwks;
     }
 
-    protected boolean isHttpsJwksInitialized(String keyLocation) throws IOException {
+    protected Get getHttpGet() {
+        return new Get();
+    }
+
+    protected boolean isHttpsJwksInitialized(String keyLocation)
+            throws IOException {
         if (mayBeFormat(KeyFormat.JWK) && keyLocation != null
                 && (keyLocation.startsWith(HTTPS_SCHEME) || keyLocation.startsWith(HTTP_SCHEME))) {
             httpsJwks = initializeHttpsJwks(keyLocation);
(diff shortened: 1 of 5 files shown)


implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/AbstractKeyLocationResolver.java
```diff
@@ -85,17 +91,31 @@ protected static String getKid(JsonWebStructure jws) {
...
+        if (location.startsWith(HTTPS_SCHEME)) {
+            Get httpGet = new Get();
+            if (tlsCertificatePath != null) {
+                httpGet.setTrustedCertificates(loadPEMCertificate(readKeyContent(tlsCertificatePath)));
+            }
+            if (trustedHosts != null) {
+                httpGet.setHostnameVerifier(new TrustedHostsHostnameVerifier(trustedHosts));
+            } else {
+                httpGet.setHostnameVerifier(new TrustAllHostnameVerifier());
+            }
```
MikeEdgar, line 108: Should the default be more defensive than trust all? I'm not sure :-)

MikeEdgar (approved):
LGTM, just one question.

implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/AbstractKeyLocationResolver.java
sberyozkin: Hi @MikeEdgar Yeah, good point, I think I'll need to introduce one more boolean property, `smallrye.jwt.tls.trust-all`, `false` is default

sberyozkin: Mike, @MikeEdgar It should be better now, have a look please. Perhaps it can make sense for you to verify that the workaround you had to use is no longer necessary with this PR ?

kdubb: I would expect this to verify hostnames by default but not require a restricted set of hostnames, unless trust-all is enabled.

MikeEdgar: > Perhaps it can make sense for you to verify that the workaround you had to use is no longer necessary with this PR ?

LGTM - I'll try it as a replacement for the workaround probably early next week.

sberyozkin: Thanks @MikeEdgar

sberyozkin:
@MikeEdgar I've simplified it quite a bit - as there was no need to update the `initializeHttpsJwks` signature with the need to update the code everywhere, and also added some basic mock tests, so it gets green then I'll merge, thanks

kdubb:
@sberyozkin Going by the documentation (and examining the code) with this PR there is no way to trust DNS host names, correct? It seems I can only disable host name checking (e.g. `trust-all`) or provide a set of allowed host names. Why not allow standard TLS semantics of checking DNS hostnames?

kdubb:
In our K8s environment we have TLS certs issued with valid CN’s and SAN’s matching DNS. This PR needlessly adds extra configuration overhead (explicitly setting hostnames) instead of relying on the standard TLS behavior.

PR #584, opened 2022-04-17
Support for HTTP Proxy

### LINKED ISSUE #460: Support for Proxy Host/Port/Auth in Key Location
Hi all,

Currently leveraging the smallrye-jwt extension in Quarkus. We have an issue in that some of our hosting environments are behind various corporate proxies and require a proxy auth before the request to the external JWT location can be made. Of course, we can simply set the JVM arguments (-Dhttp.proxyHost,Port,etc) inside our image, but we'd like to restrict the proxy access to this specific refresh command for security reasons.

It looks like the jose4j lib has direct support for setting proxy on connections [here](https://bitbucket.org/b_c/jose4j/src/f6655ef41cff1737b52c8ba9285e819843a94b37/src/main/java/org/jose4j/http/Get.java#lines-339).

So I think it would just involve adding a few proxy config properties (hostname/username/password/port/enabled ?), creating a proxy object and passing that to the httpsjwk connection so that refreshes go through the proxy first. I'd be willing to take a crack at it if you agree there's a need.

Let me know what you think -- thanks!

sberyozkin:
Fixes #460

`Jose4J`'s `Get` class accepts `Proxy` (no control over the authentication though), so it makes sense to fix this long pending issue  since we have just added the config for TLS certs/hosts too.

Default proxy host is 80, same as it is done for Vertx

### DIFF
--- a/doc/modules/ROOT/pages/configuration.adoc
+++ b/doc/modules/ROOT/pages/configuration.adoc
@@ -64,6 +64,8 @@ SmallRye JWT supports many properties which can be used to customize the token p
 |smallrye.jwt.tls.certificate.path|none|Path to TLS trusted certificate which may need to be configured if the keys have to be fetched over `HTTPS`.
 |smallrye.jwt.tls.trust-all|false|Trust all the hostnames. If the keys have to be fetched over `HTTPS` and this property is set to `true` then all the hostnames are trusted by default.
 |smallrye.jwt.tls.trusted.hosts|none|Set of trusted hostnames. If the keys have to be fetched over `HTTPS` and `smallrye.jwt.tls.trust-all` is set to `false` then then this property can be used to configure the trusted hostnames.
+|smallrye.jwt.http.proxy.host|none|HTTP proxy host.
+|smallrye.jwt.http.proxy.port|80|HTTP proxy port.
 |===
 
 = Create JsonWebToken with JWTParser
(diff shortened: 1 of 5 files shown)


MikeEdgar (commented):
@sberyozkin this looks good. My only question that I didn't think of for the last PR is - should the new properties (this PR and #582) be in the `smallrye.jwt.jwks` namespace instead? It seems that might make their purpose more obvious when just looking at the property name.

sberyozkin:
Hi Mike @MikeEdgar Sorry, I missed your comment, IMHO we should avoid `.jwks.` as, even though fetching JWKs is the main case where these properties can be applied, we can also have PEM certs or keys fetched remotely, and `smallrye-jwt` will end up supporting the remote token introspection then these properties will also apply.

sberyozkin:
@MikeEdgar Yeah, you are right the code is restricted to HttpsJwks at the moment which can only be used to fetch JWKs, but I'll need to follow later with another PR when fetching PEM keys/certs.

However, what I guess what needs to be clarified is that it really an HTTP client which is being configured here, so
I guess I'll just update the docs, smallrye-jwt itself does not have any HTTP server support
, while adding `client` like this:
`smallrye.jwt.tls.client.certificate-path`
`smallrye.jwt.tls.client.trust-all`

may be not necessary as there will never be `tls.server` at the smallrye-jwt level, while `http.proxy` is implicitly client specific

Does it sound reasonable ?

MikeEdgar:
@sberyozkin that sounds good to me. My only thought/question is - should it be `client.tls` rather than `tls.client`? No strong preference for me, but it seems more conventional.

sberyozkin:
@MikeEdgar Sure, I'll open another PR shortly to have `.client.tls.`. if you are Ok then with this particular PR then please approve and I'll merge

MikeEdgar (approved):
:+1: 
Thanks @sberyozkin

sberyozkin:
Thanks Mike :-)

PR #586, opened 2022-04-19
Use smallrye.jwt.client.tls namespace for configuring TLS

sberyozkin:
Mike @MikeEdgar has proposed that a `client` qualifier is used in the configuration namespace for TLS, it indeed makes it clearer that they only affect the (http) client (currently when fetching JWKs, to be improved for fetching PEM keys too)


MikeEdgar (approved):
LGTM, thank you

sberyozkin:
Thanks @MikeEdgar

Issue #593, opened 2022-05-02
Support for `data:` URLs for loading resources


MikeEdgar:
Currently, the loading of resources is limited to `file`, `https`, and `classpath` URLs [1]. It would be useful to also support `data` URLs [2] as well. In a cloud environment, this would allow us to directly specify resources via env (from secrets) instead of mounting the secrets as volumes and mapping to files.

For example, the configuration could be set to:
```
smallrye.jwt.client.tls.certificate.path=data:,${my.pem.encoded.ca-cert}
```
Allowing the secret to be mapped to env `MY_PEM_ENCODED_CA_CERT` and loaded by the implementation.

[1] https://github.com/smallrye/smallrye-jwt/blob/main/implementation/common/src/main/java/io/smallrye/jwt/util/ResourceUtils.java#L64
[2] https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs

sberyozkin:
if you agree with the `smallrye.jwt.client.tls.certificate` idea then we can still track the option of supporting custom URL schemes with this issue

MikeEdgar:
@sberyozkin I think the suggestion to have `smallrye.jwt.client.tls.certificate` is good. I spent a little time looking into the `data:` URL and there is actually some complexity around the payload being URL-encoded which reveals it may not be viable for this use case.

sberyozkin:
Hey @MikeEdgar Do you think we can keep this issue as a possible enhancement ?

MikeEdgar:
@sberyozkin , no objection from me to keep it open. There could be cases where it's useful, although #596 was exactly what I needed at the time :smile:

sberyozkin:
Hey @MikeEdgar, should we close this issue, np if you prefer to keep open, I'm not sure if/when I'll  be looking into it :-)

MikeEdgar:
@sberyozkin let's close it, no need to keep it at this point.

sberyozkin:
Thanks @MikeEdgar

PR #596, opened 2022-05-16
Add smallrye.jwt.client.tls.certificate to support inlined TLS certificates

### LINKED ISSUE #595: Support for inlined TLS certificates
We agreed in #593 that it can be useful in some cases to let users inline a trusted certificate in the configuration.

sberyozkin:
Fixes #595.

@MikeEdgar Hi Mike, I'm keeping #593 open, in case we'd like to go ahead and support random URL schems via the custom resolvers.

I've updated one of the unit tests to check a new property since there is another test there which checks the certificate path


MikeEdgar (approved):
This looks good, thank you @sberyozkin

PR #597, opened 2022-05-17
Support for keystores

### LINKED ISSUE #85: Support for KeyStores for keeping the keys

sberyozkin:
Fixes #85

This PR provides an initial support for KeyStores. If either `mp.jwt.verify.publickey.location` or `mp.jwt.decrypt.key.location` or both of these properties point to a `Keystore` then the following new properties can be used to extract the keys and customize the way KeyStore is loaded:
* `smallrye.jwt.keystore.password` - KeyStore password
* `smallrye.jwt.keystore.type` - optional type, example, `PKCS11`, default is `JKS`
* `smallrye.jwt.keystore.provider` - optional provider (ex, BC FIPS, etc)
* `smallrye.jwt.keystore.verify.key.alias` - this is a verification public key alias, will be extracted from a matching certificate
* `smallrye.jwt.keystore.decrypt.key.alias` - this is a decryption private key alias
* `smallrye.jwt.keystore.decrypt.key.password` - optional decryption private key password if it is different from a keystore password

Right now, if both  `mp.jwt.verify.publickey.location` and `mp.jwt.decrypt.key.location` point to a Keystore then they must point to the same store, since a keystore can have many entries. I did not want to start supporting different sets of properties for KeyStores keeping verification keys only and KeyStores keeping the descryption keys only - I'm not sure it will ever be necessary. However, it will be possible to migrate to such a model seamlessly if needed: `smallrye.jwt.keystore.password`,  `smallrye.jwt.keystore.type`, `smallrye.jwt.keystore.provider` will become the fallback properties and `verify` and `decrypt` specific properties will be introduced and checked before the fallbacks. But I don't think it is needed right now.

Also, I did not want to support the cases yet where a token `kid` will act as a key alias, that might be done in the future.

Additionally, I've updated the JWT Build code to support keystores for the `sign()`/`encrypt()/innersign()` cases, with two more properties:

* `smallrye.jwt.keystore.encrypt.key.alias` - this is an encryption public key alias, will be extracted from a matching certificate
* `smallrye.jwt.keystore.sign.key.alias` - this is a signing private key alias
* `smallrye.jwt.keystore.sign.key.password` - optional signing private key password if it is different from a keystore password

@darranl Darran, you have also mention in #85 that Elytron supports `KeyStore which can wrap another KeyStore to filter the result` - it can be interesting as the reason I've finally started looking at this issue is because there is a requirement for `smallrye-jwt` to work with `PKCS11` stores


sberyozkin:
Hey @MikeEdgar Thanks, please take your time, early next week will be fine as well I guess, cheers

sberyozkin:
Both the `server` and `jwt-build` use the same properties to refer to the actual keystore file (and password), provider, type. Right now I'm not sure it is worth supporting 4 different keystore locations (one for signing, one for decryption, and the same on the receiving end), single keystore, ex on RHEL FIPS will do, but we can diversify later if needed

testsuite/basic/src/test/java/io/smallrye/jwt/auth/principal/KeyStoreLocationTest.java
```diff
@@ -0,0 +1,62 @@
+/*
+ *   Copyright 2019 Red Hat, Inc, and individual contributors.
```
MikeEdgar, line 2: Old year here

implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtSignTest.java
```diff
@@ -573,6 +575,31 @@ public void testSignClaimsSymmetricKey() throws Exception {
...
+    public void testSignWithKeyStore() throws Exception {
+        JwtBuildConfigSource configSource = getConfigSource();
+        configSource.setUseKeyStore(true);
+        configSource.setSigningKeyLocation("/keystore.jks");
+        try {
+            String jwt = Jwt.claims()
+                    .claim("customClaim", "custom-value")
+                    .sign();
+
+            KeyStore keyStore = KeyUtils.loadKeyStore("keystore.jks", "password", Optional.empty(), Optional.empty());
```
MikeEdgar, line 588: Might be good to make this new keystore a PKCS12 file to hit the alternate code branch.

MikeEdgar (approved):
Looks good, just a couple comments.

sberyozkin:
@MikeEdgar Thanks for the review, will address the comments soon

implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtSignTest.java
sberyozkin: @MikeEdgar I've changed it to use a `PKCS12` keystore type, which can be confirmed with `keytool -list -storetype PKCS12 -keystore keystore.p12`, but at the test level I could not cause a failure to prove it is a PKCS12 store by reading it as JKS because by default it just works, 
```
KeyUtils.loadKeyStore("keystore.p12", "password", Optional.of("JKS"), Optional.empty());
```

due to the compatibility mode, and setting a system property is needed to disable it, https://www.oracle.com/java/technologies/javase/8u60-relnotes.html...

sberyozkin:
@MikeEdgar Let me go ahead with the merge now, comments have been addressed, thanks

PR #605, opened 2022-06-08
Support for an inlined decryption key

sberyozkin:
This simple PR makes sure the decryption keys can be inlined directly in the configuration the same way it can be done for the verification keys - the motivation is to support a simple extraction of both verification and decryption keys from DB and other sources via `ConfigSource` as one of the users can not verify the inner signed tokens with the keys available in DB. The same was done earlier for `smallrye-jwt-build`. 
Perhaps support for custom KeyLocationResolvers can also be introduced in the next stage, I think Mike tried it earlier but I was concerned about exposing Jose4j specific details


sberyozkin:
Also added a `Deprecated` annotation to `smallrye.jwt.decrypt.key.location` as it was deprecated long time ago but I forgot to mark it.

MikeEdgar (approved):
LGTM

sberyozkin:
Thanks Mike

PR #606, opened 2022-06-09
Support for reading from default Keystore

sberyozkin:
Minor follow up to the keystore related code: if both the inlined key is null and the key location is null but the keystore alias (and password for the private keys) are set then attempt to load from a default store. It is to support this code:

https://docs.oracle.com/en/java/javase/17/security/pkcs11-reference-guide1.html

Right now one needs to point to the existing keystore location and it is not obvious reading the NSSDB as an InputStream and loading KeyStore from it will work or is sound as this DB may be large

CC @pjgg

### DIFF
--- a/implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/util/KeyUtils.java
@@ -285,8 +285,12 @@ public final class KeyUtils {
         KeyStore keyStore = keyStoreProvider.isPresent()
                 ? KeyStore.getInstance(theKeyStoreType, keyStoreProvider.get())
                 : KeyStore.getInstance(theKeyStoreType);
-        try (InputStream is = ResourceUtils.getResourceStream(keyStorePath)) {
-            keyStore.load(is, keyStorePassword.toCharArray());
+        if (keyStorePath != null) {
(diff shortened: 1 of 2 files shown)


sberyozkin:
@MikeEdgar @radcortez sorry, I forgot we discussed with @pjgg this option so I'd like to try and squeeze it into 3.5.1 too - right now even though keystore are supported one can't request loading from a default one

MikeEdgar (approved):
It this functionality dependent on a certain provider? Typically using load with null input with give an empty keystone, right?

sberyozkin:
@MikeEdgar Hi Mike, yes, loading a null `PKCS11` type keystore will get a `SunPKCS11` provider used to load NSSDB on RHEL FIPS, as per the linked docs, sorry I see the link was incomplete, https://docs.oracle.com/en/java/javase/17/security/pkcs11-reference-guide1.html#GUID-4C366313-33B9-458C-A845-33D0C8A9C367

MikeEdgar:
> loading a null `PKCS11` type keystore will get a `SunPKCS11` provider used to load NSSDB on RHEL FIPS, as per the linked docs, sorry I see the link was incomplete

Thanks @sberyozkin , I skimmed the doc, but didn't catch that :+1:

sberyozkin:
Np at all :-)

PR #616, opened 2022-07-03
Allow for non-RSA/EC PrivateKeys when building JWT

sberyozkin:
This is a minor update to the way the keys are checked when `smallrye-jwt-build` runs on RHEL FIPS. In most cases a generic `P11Key$P11PrivateKey` will be returned which can only be used if `SunPKCS11` provider is installed. At the moment `smallrye-jwt-build` fails early in such cases which is a problem


MikeEdgar (approved):

sberyozkin:
Thanks @MikeEdgar @radcortez

PR #619, opened 2022-07-11
Use a Provider instance instead of a provider string in KeyStore.getInstance

sberyozkin:
CC @pjgg

This PR updates the code which loads `Keystore` to use a `Provider` instance instead of a string, and also adds as an option to load from a `null` (default) KeyStore to `smallrye-jwt-build` (as a follow up to #606 which missed it and only added this support for the verification part)


sberyozkin:
Hey @MikeEdgar @radcortez Can you approve this one please as Pablo has confirmed that it finally started working on RHEL FIPS ? I'll follow up with a `3.5.3` release

MikeEdgar (approved):
:+1:

sberyozkin:
Thanks @MikeEdgar

PR #638, opened 2022-09-07
Provide cause with signingKeyCanNotBeLoadedFromLocation

### LINKED ISSUE #631: Provide cause with `signingKeyCanNotBeLoadedFromLocation`
The `catch` block in `io.smallrye.jwt.build.impl.JwtSignatureImpl#getKeyContentFromLocation` doesn't include the cause into the `signingKeyCanNotBeLoadedFromLocation` exception it throws, although that detail can be very helpful.

This is such a minor change that I don't want to jump through the fork-pr-loop ;-)

sberyozkin:
Fixes #631


sberyozkin:
Let me merge this one, it is simple enough for me not to break anything :-)

PR #640, opened 2022-09-07
Fail loading the keystore if no configured provider is found

sberyozkin:
Defaulting to a null provider if the configured/requested provider is not found makes it more difficult to debug 
CC @pjgg


Ladicek:
There's a branch protection rule that requires a check named "build with jdk 8" to pass. Go to https://github.com/smallrye/smallrye-jwt/settings/branches and click _Edit_ on the branch protection rule for `main`.

radcortez:
Exactly. I guess I forgot to remove the check.

sberyozkin:
thanks @Ladicek, it helped

radcortez (approved):

MikeEdgar (approved):

PR #662, opened 2022-12-21
Add support for EDDSA

sberyozkin:
This PR adds a support to the EDDSA family of signature and encryption algorithms now also supported at the jose4j level.
EDDSA and its Ed25519 variant is considered to be one of the most secure and compact modern signature algorithms.

Good overview is for ex here:
https://cryptobook.nakov.com/digital-signatures/eddsa-and-ed25519

FYI, GitHub now uses Ed25519 with SSH keys: https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent#generating-a-new-ssh-key

I'd like to release smallrye-jwt once this PR is merged

### DIFF
--- a/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
@@ -6,21 +6,28 @@ package io.smallrye.jwt.algorithm;
  * @see <a href="https://tools.ietf.org/html/rfc7518#section-3">https://tools.ietf.org/html/rfc7518#section-3</a>
  */
 public enum SignatureAlgorithm {
-    RS256,
-    RS384,
-    RS512,
-    ES256,
-    ES384,
-    ES512,
-    HS256,
-    HS384,
-    HS512,
-    PS256,
-    PS384,
-    PS512;
+    RS256("RS256"),
+    RS384("RS384"),
+    RS512("RS512"),
+    ES256("ES256"),
+    ES384("ES384"),
+    ES512("ES512"),
+    EDDSA("EdDSA"),
+    HS256("HS256"),
+    HS384("HS384"),
+    HS512("HS512"),
+    PS256("PS256"),
+    PS384("PS384"),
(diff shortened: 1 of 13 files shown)


sberyozkin:
Ouch. It had to be excluded for Java 11. I'll think what can be done

sberyozkin:
@radcortez @MikeEdgar I restricted the new tests to `Java major version >=17` while in the actual code, where the checks against concrete Java17+ only interfaces was made, I replaced them by checking the interface name which, while Java11 is still supported, is not too bad - these are light weight checks, and they are only performed if a user did not set up `EdDSA` in the builder API or when using a parser manually to verify/decrypt - not configured it to use `EdDSA`.
Postponing this update for another year or so would not be ideal given it is a very popular algorithm now.

MikeEdgar (approved):
LGTM @sberyozkin . My only comment/suggestion would be to consolidate the isXXXKey methods with an added parameter for the class name being checked. Also, is there any case where you'd need to guard against a null key?

radcortez (dismissed):

sberyozkin:
Thanks @MikeEdgar @radcortez.

Mike,
>My only comment/suggestion would be to consolidate the isXXXKey methods with an added parameter for the class name being checked.

I can create KeyUtils methods like  `isSupportedKey(Key key, String intfName)` and have the current ones delegate to it...The current method names are mode descriptive so I'd not mind to keep them but indeed the interface checking code can be moved to KeyUtils.

> Also, is there any case where you'd need to guard against a null key?

A method like `sign("/mykey.pem")` will fail if it fails to load the key, so the only NPE case is when someone calls `sign((PrivateKey)null)`, but I'm not sure we need to protect against such cases. That said, it is a simple check, I might as well add it.

I'll push an update shortly which will reset your approvals, but I'll go ahead once it is green

sberyozkin:
@MikeEdgar I moved the the interface name check code to KeyUtils, added NPE checks at the smallrye-jwt-build for now as well

MikeEdgar (approved):
LGTM, thanks @sberyozkin

PR #669, opened 2023-01-23
Jakarta and main branch sync

sberyozkin:
I've cherry-picked all commits relevant for the Jakarta branch since 
https://github.com/smallrye/smallrye-jwt/commit/f15979e4b28235946be4d03b6bc91d02d7d90765

However, Roberto, is there a better Github command sequence that you can recommend. I don't really expect something has been missed, but just in case. Right now we have no all those release smallrye-jwt 3.x commits which is probably better but not sure it is 100% correct


radcortez (approved):

sberyozkin:
I'll follow the same strategy as the one used by smallrye-config just to avoid any risks of missing something, and then will be individually merging to the jakarta branch

PR #693, opened 2023-03-23
Add support for mp.jwt.verify.clock.skew property

Skyllarr:


implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
```diff
@@ -139,7 +138,11 @@ private JwtContext parseClaims(String token, JWTAuthContextInfo authContextInfo,
                 builder.setExpectedIssuer(authContextInfo.getIssuedBy());
             }
 
-            if (authContextInfo.getExpGracePeriodSecs() > 0) {
+            // mp.jwt.verify.clock.skew property has precedence over smallrye.jwt.expiration.grace
+            // smallrye.jwt.expiration.grace has default 60 seconds which is used if none of the properties are present
+            if (authContextInfo.getClockSkew() != null && authContextInfo.getClockSkew() > 0) {
+                builder.setAllowedClockSkewInSeconds(authContextInfo.getClockSkew().intValue());
+            } else if (authContextInfo.getExpGracePeriodSecs() > 0) {
```
Skyllarr, line 145: @sberyozkin Let me know if you agree with usage of _Long_ for this property and if you agree with clockSkew having precedence and _expiration.grace_ being ignored if _clock.skew_ is present. Thank you!

sberyozkin: Hi @Skyllarr Thanks for the PR, should we have this property set to `60` by default ?

implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -710,6 +722,7 @@ Optional<JWTAuthContextInfo> getOptionalContextInfo() {
         contextInfo.setHttpProxyPort(httpProxyPort);
         SmallryeJwtUtils.setContextGroupsPath(contextInfo, groupsPath);
         contextInfo.setExpGracePeriodSecs(expGracePeriodSecs);
+        contextInfo.setClockSkew(mpJwtVerifyClockSkew.orElse(null));
```
sberyozkin, line 725: Please remove a default value for `expGracePeriodSecs` and it is set to something - log that this property is deprecated and that a new one should be used, similarly to how done for other deprecated properties.
The new property should be set to `60` by default to retain the same expectations after  `expGracePeriodSecs` is no longer initialized to 60 by default.
Does it sound reasonable ?

sberyozkin:
@Skyllarr Thanks very much for the PR, can you also please update the configuration doc, list a new property in the table listing standard properties and update the smallrye jwt property description to say it is deprecated and that a new property should be used, you can copy and paste a similar text for some other properties there

implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTTokenParser.java
Skyllarr: @sberyozkin I did not do it at fist since I have seen [this older commit](https://github.com/eclipse/microprofile-jwt-auth/pull/286/files) , but yes given the situation changed since it sounds good. I have updated the PR accordingly

implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
Skyllarr: @sberyozkin Yes, thank you! Updated

Skyllarr:
> @Skyllarr Thanks very much for the PR, can you also please update the configuration doc, list a new property in the table listing standard properties and update the smallrye jwt property description to say it is deprecated and that a new property should be used, you can copy and paste a similar text for some other properties there

@sberyozkin Thank you for feedback! I've updated the configuration doc and updated the property descriptions. Let me know if there is anything else missing. Thanks!

sberyozkin (approved):
Thanks @Skyllarr

PR #707, opened 2023-06-27
Support AWS ALB token verification

### LINKED ISSUE #708: Support AWS ALB token verification
See https://github.com/quarkusio/quarkus/issues/34300

sberyozkin:
Fixes #708.

This PR adds a new property `smallrye.jwt.verify.key-provider` which can be set to `AWS_ALB` - it will register a verification key resolver appending the token kid to the public key location property and fetch the key. 

Typical configuration to verify AWS ALB tokens:

```
mp.jwt.verify.publickey.location=${aws.alb.key.endpoint}
mp.jwt.verify.issuer=${aws.alb.token.issuer}
mp.jwt.verify.publickey.algorithm=ES256

smallrye.jwt.verify.key-provider=AWS_ALB
smallrye.jwt.time-to-live=-1
```

`iat` (issued at claim) is required in MP-JWT but AWS ALB token does not have an `iat` claim set, so `smallrye.jwt.time-to-live=-1` relaxes this requirement indirectly since without `iat` the time to live time can not be calculated.

Note the token parts have been `Base64` encoded, as opposed to `Base64Url` encoded, thus extra padding characters are added to the header and body and signature parts. 

As far as the the header and body integrity verification is concerned, it makes no difference because the padding characters are included in the signing input - while for the signature it does not matter if its encoded form has extra padding characters or not - since the decoded signature is compared against the recalculated signature with the the headers and body parts included in the signing input.

Base64-only encoded tokens introduce an interoperability problem - i.e not all libraries can consume such tokens - but it is not a problem of Jose4j - rather it is a problem of AWS ALB whose tokens may not be verified by all JOSE libraries.

Such tokens are also not URI friendly, as `=` may get URL encoded.


sberyozkin:
Hey @radcortez @AdamBien, 

I've added the basic cache support (similar to what we have in Quarkus OIDC for caching UserInfo/TokenIntrospection).

I've been contemplating for a while how to configure this in-mem cache. I thought it may not be even necessary since the key rotation is not happening every minute, the cache size will probably reach 10 max for a lifetime of a given pod in OpenShift.
But there there would always be a concern, so I looked at adding Map based configuration, Roberto advised the use config mapping but while trying to make it work I thought that may be using Map would not add any real benefit - as the Map properties would be untyped, we'd need to doc them for AWS ALB resolver, then the next similar comes arrives - duplicate the same, so I just thought having 2 generic key cache size and time-to-live properties would cover most of the variations going for a while whenever a dynamic key resolver is activated.

AdamBien:
> Hey @radcortez @AdamBien,
> 
> I've added the basic cache support (similar to what we have in Quarkus OIDC for caching UserInfo/TokenIntrospection).
> 
> I've been contemplating for a while how to configure this in-mem cache. I thought it may not be even necessary since the key rotation is not happening every minute, the cache size will probably reach 10 max for a lifetime of a given pod in OpenShift. But there there would always be a concern, so I looked at adding Map based configuration, Roberto advised the use config mapping but while trying to make it work I thought that may be using Map would not add any real benefit - as the Map properties would be untyped, we'd need to doc them for AWS ALB resolver, then the next similar comes arrives - duplicate the same, so I just thought having 2 generic key cache size and time-to-live properties would cover most of the variations going for a while whenever a dynamic key resolver is activated.

The key rotation happens on hourly bases. I was able to test for several hours without any key changes. Because the key is announced in the kid, the key value could be cached as long as it does not change. Do we need any configuration here?

Regarding  "a given pod in OpenShift." Quarkus as AWS Lambda can also be deployed behind an ALB.  There are two use cases / target platforms for this extension: Quarkus as AWS Lambda and Quarkus running in a container (EKS, ECS, ...)

sberyozkin:
@AdamBien 

>The key rotation happens on hourly bases. I was able to test for several hours without any key changes. Because the key is announced in the kid, the key value could be cached as long as it does not change. Do we need any configuration here?

I'm not sure  right now we need more configuration related to the cache control, we have the max size, and the time to live. Having a timer to run a background task to clean it may be an option but it is easier do at the Quarkus level, probably not essential at the moment as the cache size is likely to remain low. 

I believe there could be be several kids active at the same time, the key may be rotated after some requests have already gone through, so these requests should be supported with the cached key.

Please clarify what you mean re the additional configuration if I missed it

Thanks

sberyozkin:
Thanks @radcortez 

By the way, I forgot to clarify why I chose to have `smallrye.jwt.verify.key-provider`, note `.verify.`, as for other verification related options, but no `.verify.` in `smallrye.jwt.key-cache-size`: this is because we have decryption key resolvers as well, so if there is ever a case of a dynamic secret key resolution requiring the local cache, we can use the same property like  `smallrye.jwt.key-cache-size`, I thought creating ` smallrye.jwt.verify.key-cache-size` and then possibly `smallrye.jwt.decrypt.key-cache-size` would be a bit too much

PR #729, opened 2023-10-13
Support for the scope claim in JWT Build API

### LINKED ISSUE #725: Add `scope` method to JWT Build API
It should be easy to set a standard scope claim, instead of `.claim("scope", "read write")` it would be easier to do `.scope("a")` or `scopes(Set.of("a", "b"))` with the space between "a" and "b" added internally

sberyozkin:
Fixes #725 

This PR provides a typed support for a `scope` claim whose origins are in OAuth2 but is used by OIDC providers.

For example, instead of writing a code like this:

https://github.com/quarkusio/quarkus/blob/main/integration-tests/oidc-tenancy/src/test/java/io/quarkus/it/keycloak/ServicePublicKeyTestCase.java#L28

I'd like to write

`Jwt.scope("read:data")`, etc.

This PR also simplifies a bit the implementation of `groups` setters

### DIFF
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/Jwt.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/Jwt.java
@@ -188,6 +188,26 @@ public final class Jwt {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Skyllarr (approved):

Skyllarr:
LGTM

PR #730, opened 2023-10-13
Support parsing already verified tokens

sberyozkin:
I've had 2 user requests over may be the last 6 months where users would like to use `JsonWebToken` API but without doing another round of verification (ex in various AWS setups) but also without registering custom factories which can be used such purposes.

A simple `parseOnly` method is added to `JwtParser` to support such cases


Skyllarr (approved):

Skyllarr:
LGTM

sberyozkin:
Thanks @Skyllarr for having a look at this and other 2 PRs

PR #731, opened 2023-10-13
Explain in JWTCallerPrincipal#getName JavaDocs that this method may not return a unique principal name

### LINKED ISSUE #714: Add a warning to the docs that `JsonWebToken.getName` is not guaranteed to be unique
See https://github.com/eclipse/microprofile-jwt-auth/issues/319.

It is important to highlight that currently, the principal name is not guaranteed to be unique and recommend using `upn`->`sub`, alongside the issuer, if the uniqueness is required.

sberyozkin:
Fixes #714


sberyozkin:
@Skyllarr, good point, right now I'm not sure how https://github.com/eclipse/microprofile-jwt-auth/issues/319 will be resolved, we may indeed end up adding the same kind of caution there, though a more energetic fix would be to remove a `preferred_username` check if the major MP JWT version - the problem is, it is hard to know when the major version will happen

Skyllarr:
@ sberyozkin makes sense, thank you!

Issue #734, opened 2023-10-23
Add KeyEncryptionAlgorithm "dir" enum value


vickymicky:
At the moment, there is no support for, key encryption algorithm -> "Direct use of a shared symmetric key as the CEK" (dir)
https://datatracker.ietf.org/doc/html/rfc7518#section-4.7

sberyozkin:
I agree, I've been contemplating to introduce it after noticing Auth0 access tokens, when encrypted, use the `dir` algorithm

PR #735, opened 2023-10-23
Support "Direct encryption with shared symmetric key"

vickymicky:
Fixes [734](https://github.com/smallrye/smallrye-jwt/issues/734)

### DIFF
--- a/implementation/common/src/main/java/io/smallrye/jwt/algorithm/KeyEncryptionAlgorithm.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/algorithm/KeyEncryptionAlgorithm.java
@@ -20,7 +20,8 @@ public enum KeyEncryptionAlgorithm {
     A256GCMKW("A256GCMKW"),
     PBES2_HS256_A128KW("PBES2-HS256+A128KW"),
     PBES2_HS384_A192KW("PBES2-HS384+A192KW"),
-    PBES2_HS512_A256KW("PBES2-HS512+A256KW");
+    PBES2_HS512_A256KW("PBES2-HS512+A256KW"),
+    DIR("dir");
 
     private String algorithmName;
 
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtEncryptionImpl.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtEncryptionImpl.java
@@ -200,9 +200,6 @@ class JwtEncryptionImpl implements JwtEncryptionBuilder {
                 throw ImplMessages.msg.unsupportedKeyEncryptionAlgorithm(alg);
             }
         }
-        if ("dir".equals(alg)) {
-            throw ImplMessages.msg.directContentEncryptionUnsupported();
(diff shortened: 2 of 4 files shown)


implementation/common/src/main/java/io/smallrye/jwt/algorithm/KeyEncryptionAlgorithm.java
```diff
@@ -20,7 +20,8 @@ public enum KeyEncryptionAlgorithm {
     A256GCMKW("A256GCMKW"),
     PBES2_HS256_A128KW("PBES2-HS256+A128KW"),
     PBES2_HS384_A192KW("PBES2-HS384+A192KW"),
-    PBES2_HS512_A256KW("PBES2-HS512+A256KW");
+    PBES2_HS512_A256KW("PBES2-HS512+A256KW"),
+    DIRECT("dir");
```
sberyozkin, line 24: I've been trying to keep the enum values matching the algorithm identifiers, so please update that to `DIR`

implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtEncryptionImpl.java
```diff
@@ -200,10 +200,7 @@ private String getKeyEncryptionAlgorithm(Key keyEncryptionKey) {
                 throw ImplMessages.msg.unsupportedKeyEncryptionAlgorithm(alg);
             }
         }
-        if ("dir".equals(alg)) {
-            throw ImplMessages.msg.directContentEncryptionUnsupported();
```
sberyozkin, line 204: Remove the corresponding message too

implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtEncryptTest.java
```diff
@@ -434,6 +434,23 @@ void encryptWithSecret() throws Exception {
...
     }
 
+    @Test
+    void encryptWithSecretKeyAndUseDirAlgo() throws Exception {
+        String jweCompact = Jwt.claims()
+                .claim("customClaim", "custom-value")
+                .jwe()
+                .keyId("key-enc-key-id")
+                .keyAlgorithm(KeyEncryptionAlgorithm.DIRECT)
+                .encrypt(createSecretKey());
```
sberyozkin, line 444: Right now it does work because the secret key is created from the same sequence, but might change. Please init a secret key var in the test and reuse it when decrypting

implementation/common/src/main/java/io/smallrye/jwt/algorithm/KeyEncryptionAlgorithm.java
vickymicky: I have fixed it

implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtEncryptTest.java
vickymicky: I have fixed it

PR #767, opened 2024-02-29
Prioritize groups path over groups claim

### LINKED ISSUE #766: Respect `smallrye.jwt.path.groups` even if the `groups` claim is present in the token
When the JWT token has a `groups` claim, it will be used, no matter what the value of `smallrye.jwt.path.groups` is.
If the groups claim contains integers (and not strings), a warning is logged and the groups list is empty.
E. g. in such cases it would be great to honor the `smallrye.jwt.path.groups` property and overwrite the actual groups.

The current implementation works as described in the docs:
> This property can be used if a token has no 'groups' claim but has the groups set in a different claim.

But I'd like to use another claim for the groups, even if groups are already present in the token. Is there any other config I'm overlooking?

oemel09:
Fixes #766
Potentially existing group claims will be overwritten by the claim specified in `smallrye.jwt.path.groups`.


sberyozkin:
@oemel09 I believe it has to be done somehow differently, something along this line: 1) if the path is set - start with checking it, if it produces nothing, check the `groups`, and finally fallback to the default role

It also should be tested

oemel09:
Question is if the `groups` should be checked at all when the claim specified by `smallrye.jwt.path.groups` is empty or if it should use the default groups (specified via `smallrye.jwt.claims.groups`) without checking the actual `groups` claim in that case.
I'd vote for the latter.

sberyozkin:
>Question is if the groups should be checked at all when the claim specified by smallrye.jwt.path.groups is empty or if it should use the default groups (specified via smallrye.jwt.claims.groups) without checking the actual groups claim in that case.

So, if the path is configured, check it, otherwise check groups, and in both cases,  fallback to the default role if it is set.
It makes sense, but I don't want to break someone's code where they may have the path configured right now which may be pointing to some non-existing claim, which has no impact anyway because the groups is always producing roles, and now, since the order is changed, and we skip the roles, their applications are broken

oemel09:
> but I don't want to break someone's code

When the `groups` claim is present in the token and the user has set `smallrye.jwt.claims.groups` to another claim which exists, the current implementation will still use the roles from the `groups` claim, but we want it to use the roles from the custom claim. So this is a breaking change already.

With the flow you described it is not possible to use the default role (via `smallrye.jwt.claims.groups`) in combination with the `smallrye.jwt.claims.groups` setting when the token also has a `groups` claim.

sberyozkin:
@oemel09 Can I ask for the clarification. Do you have tokens with the custom `groups` claim whose value has nothing to do with roles ? Or you have tokens with the `groups` claim containing roles but you want some other claim be used as a source of roles ?

oemel09:
It's the former, in the `groups` claim in the token there are integers. But what I'm interested in is stored in a `role` claim.

oemel09:
I changed the implementation in the way you suggested and also added tests to it.
As explained I'm still in favor of the other flow. But anyway, for my use case both flows would work.
What's the way forward here? And, in case you know it, how long will it take for this change to end up in quarkus? 🙈

sberyozkin:
@oemel09 I've been thinking about it for a while, and came to the conclusion that I was over-thinking about the side-effects. 
If someone does set the custom role claims path then it should simply be used.

So lets get back to what I believe you were suggesting:
* if the custom path is set - use it
* otherwise check `groups`
* If the roles is null and the default role is set - use it

Please do the final update per the above and resolve this issue

Thanks for the patience

oemel09:
I did the changes and adapted the tests. I think I added all the relevant cases, please have a look at them and check if you are okay with all the expected results for each combination.

doc/modules/ROOT/pages/configuration.adoc
```diff
@@ -56,9 +56,9 @@ SmallRye JWT supports many properties which can be used to customize the token p
...
 |smallrye.jwt.claims.sub|none| This property can be used to set a default sub claim value when the current token has no standard or custom `sub` claim available. Effectively this property can be used to customize `java.security.Principal` name if no `upn` or `preferred_username` or `sub` claim is set.
-|smallrye.jwt.path.groups|none|Path to the claim containing the groups. It starts from the top level JSON object and can contain multiple segments where each segment represents a JSON object name only, example: `realm/groups`. This property can be used if a token has no 'groups' claim but has the groups set in a different claim. Use double quotes with the namespace qualified claims.
+|smallrye.jwt.path.groups|none|Path to the claim containing the groups. It starts from the top level JSON object and can contain multiple segments where each segment represents a JSON object name only, example: `realm/groups`. This property can be used if a token the groups set in a different claim than `groups`. Use double quotes with the namespace qualified claims.
```
sberyozkin, line 59:
```suggestion
|smallrye.jwt.path.groups|none|Path to the claim containing the groups. It starts from the top level JSON object and can contain multiple segments where each segment represents a JSON object name only, example: `realm/groups`. This property can be used if a token has the groups set in a claim different to `groups`. Use double quotes with the namespace qualified claims.
```

implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/PrincipalUtils.java
```diff
@@ -46,14 +46,23 @@ public static void setClaims(JwtClaims claimsSet, String token, JWTAuthContextIn
...
+                roles = List.class.cast(groupsClaim);
+            }
+        } else {
+            roles = findGroups(authContextInfo, claimsSet);
+        }
+
+        if (roles == null && authContextInfo.getDefaultGroupsClaim() != null) {
+            roles = Collections.singletonList(authContextInfo.getDefaultGroupsClaim());
         }
+        claimsSet.setClaim(Claims.groups.name(), roles);
```
sberyozkin, line 65: Please check if it is not null (the current code on the main branch does not check it but it should)

oemel09: This would then still return the `groups` claim although the custom path is set when the set custom path is empty.
This would mean this test case must be changed: https://github.com/smallrye/smallrye-jwt/pull/767/files#diff-30eaa73c4738eaad7b0a7e988717d1a2c0e5256141e3f00a3d7cc2cc8241db83R55
If that's what's desired, I'm okay with changing it...

I think it would be easier now that the tests are there to comment on those with the expected behavior.

oemel09: I assume you're talking about the `roles` or are you talking about the `claimsSet`?

sberyozkin: @oemel09 yes,`roles`

oemel09: Okay, I adapted the implementation

sberyozkin (approved):

PR #773, opened 2024-03-21
Add JavaDocs to DefaultJWTParser constructors and prevent possible NPE

### LINKED ISSUE #765: JWTCallerPrincipalFactory is instantiated each time if not explicitly set
When the instantiation of the JWTCallerPrincipalFactory is not externally managed (i.e. when using io.smallrye.jwt.auth.principal.DefaultJWTParser#DefaultJWTParser(io.smallrye.jwt.auth.principal.JWTAuthContextInfo) ), it is instantiated each time a token is parsed.

This causes the JWK sets to never be cached and a call to the OIDC provider is always made to retrieve them. Implicitely, it has a huge impact on performances.
Just making the factory producer RequestScoped (or using a ThreadLocal when not using CDI) has a huge performance improvement, without any concurrency risk.

Was it a deliberate choice?

sberyozkin:
Fixes #765 

Hi @MikeEdgar Can you please review this is very last PR before the release ? It only adds JavaDocs as users may be thinking the factory is managed somehow when this parser is created directly in the user code. 
It also fixes a possible NPE if the constructor accepting the factory only is used. This constructor supports methods like `parse(token, myPublicKey)` where a new context is created in any case, but if this constructor is used with `parse(token)` which is meant to be used with the injected JWTParser only, then it will cause NPE, so this PR prevents it

### DIFF
--- a/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
+++ b/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
@@ -55,17 +55,39 @@ public class DefaultJWTParser implements JWTParser {
     @Inject
     private JWTCallerPrincipalFactory callerPrincipalFactory;
 
+    /**
+     * Default constructor which is required to support a JWTParser injection
+     */
     public DefaultJWTParser() {
     }
 
+    /**
+     * Constructor which initializes DefaultJWTParser with the provided {@link JWTAuthContextInfo} and a new instance of
+     * {@link JWTCallerPrincipalFactory}
+     *
+     * @param authContextInfo {@link JWTAuthContextInfo}
+     */
     public DefaultJWTParser(JWTAuthContextInfo authContextInfo) {
         this(authContextInfo, new JWTCallerPrincipalFactoryProducer().getFactory());
     }
 
+    /**
+     * Constructor which initializes DefaultJWTParser with the provided {@link JWTCallerPrincipalFactory} and a new instance of
+     * {@link JWTAuthContextInfo}
+     *
+     * @param factory {@link JWTCallerPrincipalFactory}
+     */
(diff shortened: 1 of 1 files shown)


implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
```diff
@@ -55,17 +55,39 @@ public class DefaultJWTParser implements JWTParser {
...
-        this(null, factory);
+        this(new JWTAuthContextInfo(), factory);
     }
 
+    /**
+     * Constructor which initializes DefaultJWTParser with the provided {@link JWTAuthContextInfo} and
+     * {@link JWTCallerPrincipalFactory}
+     *
+     * @param authContextInfo {@link JWTAuthContextInfo}
+     * @param authContextInfo {@link JWTCallerPrincipalFactory}
```
MikeEdgar, line 89:
```suggestion
     * @param factory {@link JWTCallerPrincipalFactory}
```

MikeEdgar (changes requested):

implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
MikeEdgar: You saw it too :)

MikeEdgar (approved):

sberyozkin:
@MikeEdgar Yes, lol, I often commit and then start checking, need to improve on it :-)

sberyozkin:
thanks @MikeEdgar

MikeEdgar:
> Yes, lol, I often commit and then start checking, need to improve on it :-)

Same here, sometimes the compare in Github is the best way to check your own work :laughing:

PR #793, opened 2024-05-04
Release SmallRye JWT 4.5.2

Eng-Fouad:


sberyozkin:
@Eng-Fouad Thanks but only smallrye-jwt committers can do releases

Eng-Fouad:
> @Eng-Fouad Thanks but only smallrye-jwt committers can do releases

I see. Sorry about that.

PR #799, opened 2024-05-28
Fix IllegalArgumentException when using EDDSA signature algorithm

0rzech:
`JwtClaimsBuilderImpl`.

to `SignatureAlgorithm.fromAlgorithm(String)`.

### DIFF
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtSignatureImpl.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtSignatureImpl.java
@@ -267,7 +267,7 @@ class JwtSignatureImpl implements JwtSignature {
 
         // Try PEM format first - default to RS256 if the algorithm is unknown
         Key key = KeyUtils.tryAsPemSigningPrivateKey(keyContent,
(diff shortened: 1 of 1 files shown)


0rzech:
Perhaps it would be better to change `SignatureAlgorithm.fromAlgorithm(String)` so that it internally calls `toUpperCase()` on alg parameter?

0rzech:
It is also possible to call `toUpperCase()` [here](https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java#L33-L35), but it would make algorithm naming lax everywyhere, as `SignatureAlgorithm.fromAlgorithm(String)` is also used [here](https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtClaimsBuilderImpl.java#L200-L220).

0rzech:
Closing in favour of #800 .

PR #800, opened 2024-05-29
Fix IllegalArgumentException when using EdDSA signature algorithm

0rzech:
`JwtClaimsBuilderImpl`.

to `SignatureAlgorithm.valueOf(String)`.

### DIFF
--- a/implementation/common/pom.xml
+++ b/implementation/common/pom.xml
@@ -57,6 +57,10 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    <dependency>
+      <groupId>net.jqwik</groupId>
+      <artifactId>jqwik</artifactId>
+    </dependency>
   </dependencies>
 
 </project>
--- a/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
@@ -22,9 +22,9 @@ public enum SignatureAlgorithm {
     PS384("PS384"),
     PS512("PS512");
 
-    private String algorithmName;
+    private final String algorithmName;
 
-    private SignatureAlgorithm(String algorithmName) {
+    SignatureAlgorithm(String algorithmName) {
         this.algorithmName = algorithmName;
     }
 
--- a/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
@@ -1,5 +1,7 @@
 package io.smallrye.jwt.algorithm;
 
+import java.util.StringJoiner;
+
 /**
  * JWT JSON Web Signature Algorithms.
  *
@@ -31,6 +33,19 @@ public enum SignatureAlgorithm {
     }
 
     public static SignatureAlgorithm fromAlgorithm(String algorithmName) {
-        return SignatureAlgorithm.valueOf(algorithmName);
+        try {
+            return SignatureAlgorithm.valueOf(algorithmName.toUpperCase());
+        } catch (Exception e) {
+            throw new IllegalArgumentException(
+                    "Invalid signature algorithm name: " + algorithmName + ", expected one of: " + getValidNames(), e);
+        }
+    }
+
+    private static String getValidNames() {
+        var names = new StringJoiner(", ");
+        for (var alg : values()) {
+            names.add(alg.getAlgorithm());
+        }
+        return names.toString();
     }
 }
--- a/implementation/common/src/test/java/io/smallrye/jwt/util/SignatureAlgorithmTest.java
+++ b/implementation/common/src/test/java/io/smallrye/jwt/util/SignatureAlgorithmTest.java
@@ -1,0 +1,71 @@
+package io.smallrye.jwt.util;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Collection;
+import java.util.HashSet;
+
+import org.junit.platform.commons.util.StringUtils;
+
+import io.smallrye.jwt.algorithm.SignatureAlgorithm;
+import net.jqwik.api.Arbitraries;
+import net.jqwik.api.Arbitrary;
+import net.jqwik.api.ForAll;
+import net.jqwik.api.Property;
+import net.jqwik.api.Provide;
+import net.jqwik.api.lifecycle.BeforeContainer;
+
+public class SignatureAlgorithmTest {
+
+    // all the names that can be used without throwing exception
+    private static final Collection<String> ACCEPTED_NAMES = new HashSet<>();
+
+    @BeforeContainer
+    static void setUp() {
+        for (var alg : SignatureAlgorithm.values()) {
+            generateAcceptedNames(alg.name().toCharArray(), 0);
+        }
+    }
+
+    @Property
+    void givenAlgorithm_shouldGetNonBlankAlgorithmName(@ForAll SignatureAlgorithm algorithm) {
+        assertTrue(StringUtils.isNotBlank(algorithm.getAlgorithm()));
+    }
+
+    @Provide
+    static Arbitrary<String> validNames() {
+        return Arbitraries.of(ACCEPTED_NAMES).dontShrink();
+    }
+
+    @Property
+    void givenValidAlgorithmName_shouldReturnAppropriateAlgorithm(@ForAll("validNames") String name) {
+        assertEquals(name.toUpperCase(), SignatureAlgorithm.fromAlgorithm(name).getAlgorithm().toUpperCase());
+    }
+
+    @Provide
+    static Arbitrary<String> invalidNames() {
+        return Arbitraries.strings()
+                .injectNull(0.0005)
+                .filter((s) -> !ACCEPTED_NAMES.contains(s));
+    }
+
+    @Property
+    void givenInvalidAlgorithmName_shouldThrowIllegalArgumentException(@ForAll("invalidNames") String name) {
+        assertThrows(IllegalArgumentException.class, () -> SignatureAlgorithm.fromAlgorithm(name));
+    }
+
+    private static void generateAcceptedNames(char[] name, int index) {
+        if (index == name.length) {
+            ACCEPTED_NAMES.add(new String(name));
+        } else if (Character.isLetter(name[index])) {
+            name[index] = Character.toUpperCase(name[index]);
+            generateAcceptedNames(name, index + 1);
+            name[index] = Character.toLowerCase(name[index]);
+            generateAcceptedNames(name, index + 1);
+        } else {
+            generateAcceptedNames(name, index + 1);
+        }
+    }
+}
--- a/implementation/common/src/test/resources/junit-platform.properties
+++ b/implementation/common/src/test/resources/junit-platform.properties
@@ -1,0 +1,1 @@
+jqwik.database=
--- a/implementation/jwt-build/pom.xml
+++ b/implementation/jwt-build/pom.xml
@@ -75,6 +75,10 @@
       <artifactId>junit-jupiter</artifactId>
     </dependency>
     <dependency>
+      <groupId>net.jqwik</groupId>
+      <artifactId>jqwik</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.parsson</groupId>
       <artifactId>parsson</artifactId>
     </dependency>
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtSignatureImpl.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtSignatureImpl.java
@@ -200,7 +200,7 @@ class JwtSignatureImpl implements JwtSignature {
             try {
                 alg = JwtBuildUtils.getConfigProperty(JwtBuildUtils.NEW_TOKEN_SIGNATURE_ALG_PROPERTY, String.class);
                 if (alg != null) {
-                    alg = SignatureAlgorithm.valueOf(alg.toUpperCase()).getAlgorithm();
+                    alg = SignatureAlgorithm.fromAlgorithm(alg).getAlgorithm();
                     headers.put(HeaderParameterNames.ALGORITHM, alg);
                 }
             } catch (Exception ex) {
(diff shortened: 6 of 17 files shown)


0rzech:
It is also possible to call `toUpperCase()` [here](https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java#L33-L35), but it would make algorithm naming lax everywyhere, as `SignatureAlgorithm.fromAlgorithm(String)` is also used [here](https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtClaimsBuilderImpl.java#L370-L376).

Alternatively, `EDDSA("EdDSA")` could be changed to `EDDSA("EDDSA")` [here](https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java#L15), because the public interface _seems to_ expect all letters to be upper case when using builder anyway, so the change should not be breaking.

sberyozkin:
@0rzech Thanks for the PR, that should have been fixed with 4.5.2, can you retry please ?

sberyozkin:
If you still see the problem, can you type the code here which leads to the correct EdDSA algorithm conversion issue ?

0rzech:
It still happens with 4.5.2 for me.

All you need to do is set

```
smallrye:
  jwt:
#    new-token:
#      signature-algorithm: EDDSA
    sign:
      key: >
        {
          "kty": "OKP",
          "crv": "Ed25519",
          "x": "9WDvp1MEM25AFpemGlPrciBA9IAYFSPNZ4xetx2v8HE",
          "d": "ogjlWAHjc3KZ1DCe8B8V7Satr6g0jpQjkGEV2lFPPIE"
        }
```

in `application.yaml` and then call

```
Jwt.claims()
    .upn("someone")
    .jws()
    .algorithm(SignatureAlgorithm.EDDSA)
    .sign();
```

in your code. Alternatively, you can remove `.algorithm(SignatureAlgorithm.EDDSA)` call from the builder step and uncomment `new-token` and `signature-algorithm` lines in yaml example.

sberyozkin:
@0rzech But the correct name is `EdDSA`, the enum value is `EDDSA` but its `.getAlgorithm` is used

0rzech:
`EdDSA` won't work either. And in fact it doesn't matter, because the name from the setting will have called `.toUpperCase()` on it first anyway [here](https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtSignatureImpl.java#L203). Also, there's an example with using `SignatureAlgorithm.EDDSA` enum directly, which should not lead to any error.

sberyozkin:
@0rzech Ok, thanks, let me have a look a little bit later just to refresh in my mind the whole logic of delaing with algo names, in general, I think I'd prefer to tune it inside `SignatureAlgorithm` if needed.

0rzech:
Sure! I can adjust PR to your decision. Unless you prefer to fix it yourself, of course. 😉

IMHO, it would be even better not to map back and forth between `String` and enum representation of `SignatureAlgorithm`, but such change would require more work.

0rzech:
I forgot to add, that the public interface requiring one to use `EDDSA` (all upper case) `String` I mentioned in one of my previous comments is [this](https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation%2Fjwt-build%2Fsrc%2Fmain%2Fjava%2Fio%2Fsmallrye%2Fjwt%2Fbuild%2Fimpl%2FJwtClaimsBuilderImpl.java#L200-L211).

0rzech:
@sberyozkin It looks like when I don't set `smallrye.jwt.token.new-token.signature-algorithm` at all and have JWK inlined in `smallrye.jwt.sign.key`, then `EdDSA` is inferred from the key and the buggy code path is not executed. I don't know if it's the same for `smallrye.jwt.sign.key.location` setting.

On the other hand, I have to set `mp.jwt.verify.publickey.algorithm` explicitly to `EDDSA` to make verification work, because key type won't be inferred from JWK in this case. Btw., `EdDSA` will result in exception here too, this time because of `java.lang.IllegalArgumentException: (...): Cannot convert EdDSA to enum class io.smallrye.jwt.algorithm.SignatureAlgorithm, allowed values: es256,es384,hs512,hs256,rs256,hs384,es512,ps512,eddsa,rs512,ps256,ps384,rs384`.

IMHO, in both these cases exception suggestions are confusing to Quarkus users, because they mention internal signature algorithm representation and the rest of stack traces' lines do not mention values expected from users at all.

sberyozkin:
Thanks for investigating @0rzech, I'll get to adding tests for different cases, it will take a bit of time, but we'll get it cleaned up

sberyozkin:
@0rzech So, lets focus on generating tokens with EdDSA first, I think we have these 2 cases working OK

1. EdDSA private key is supplied directly to `sign(PrivateKey)` - we have a test for this case
2. As you have confirmed, with the inlined JWK alone, the algorithm is correctly inferred - if I remember correctly the test for it does exist too

These 2 cases do not work:

* `.jws().algorithm(SignatureAlgorithm.EDDSA)`
* ` smallrye.jwt.new-token.signature-algorithm=EDDSA`

Can you please try to add 2 tests for these failing cases ? I'd also prefer if we can manage the correct conversion for EdDSA inside `SignatureAlgorithm` . Code like https://github.com/smallrye/smallrye-jwt/blob/7807b3f1b5db36c7ad4484a5dae4ed40230713e1/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/JwtSignatureImpl.java#L203 should be gone and replaced with `.fromAlgorithm()`.

Please take your time, I'll be on PTO and realistically I'll have time to review and merge this PR from 11th June onwards

sberyozkin:
https://github.com/smallrye/smallrye-jwt/blob/main/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java#L34 can handle it nicely.

Please also add extra log/error messages if you prefer

implementation/common/pom.xml
```diff
@@ -57,6 +57,10 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    <dependency>
```
sberyozkin, line 60: We don't need another dependency IMHO

implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
```diff
@@ -31,6 +33,19 @@ public String getAlgorithm() {
...
-        return SignatureAlgorithm.valueOf(algorithmName);
+        try {
+            return SignatureAlgorithm.valueOf(algorithmName.toUpperCase());
+        } catch (Exception e) {
+            throw new IllegalArgumentException(
+                    "Invalid signature algorithm name: " + algorithmName + ", expected one of: " + getValidNames(), e);
+        }
+    }
+
+    private static String getValidNames() {
```
sberyozkin, line 44: We only need to make sure that if someone typed `rs256` or `RS256`, it works, so this is why to upper case conversion was there, and do a single specific `if` for `EdDSA`, this kind of check does not need a new dependency IMHO

0rzech: Jqwik is used here as well just not to mix it with `Junit.Parametrized`.

The true reason for jqwik is testing the method's property of failing for any other string than the valid ones [here](https://github.com/0rzech/smallrye-jwt/blob/e52d7f0bb542550a305e0c4fbacd80e4f9966e4e/implementation/common/src/test/java/io/smallrye/jwt/util/SignatureAlgorithmTest.java#L41-L44).

implementation/common/pom.xml
0rzech: If we want to be able to do property-based tests, then we need. Given it's just a test dependency, it would be shame to resign from property-based tests. But it's your call.

implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
0rzech: Jqwik is used here as well just not to mix it with `Junit.Parametrized`.

The true reason for jqwik is testing the method's property of failing for any other string than the valid ones [here](https://github.com/0rzech/smallrye-jwt/blob/e52d7f0bb542550a305e0c4fbacd80e4f9966e4e/implementation/common/src/test/java/io/smallrye/jwt/util/SignatureAlgorithmTest.java#L41-L44) and most likely in other tests I have yet to write.

implementation/common/pom.xml
0rzech: If we want to be able to do property-based tests, then we do need it. Given it's just a test dependency, it would be shame to resign from this. But it's your call.

sberyozkin: @0rzech I see, I missed you meant to use it for tests only. These tests look fine, but for some reasons this dependency requires a dedicated exclusion in .gitignore. I appreciate you like the style offered but me being quite conservative, I'd say we can easily enough, even if it will be more verbose, to write similar tests without it.

Let me ask Mike. @MikeEdgar, do you think we should bring it or just use existing test dependencies ?

0rzech: `.jqwik-database` is a file to store data of previous runs. Given how property-based testing works with shrinking etc., I don't think we should try to emulate it ourselves.

sberyozkin: At the same time, we don't really need to introduce it IMHO if it keeps some local state after runs

0rzech: It's helpful if you want to re-run failing properties, but I can disable it.

sberyozkin:
@0rzech It certainly looks like a high quality PR, thanks a million for it. But, please don't get me wrong, this dependency makes me a bit nervous, possibly due to my ignorance,  but having to add these exclusions here and there looks like an extra hassle that can be avoided. And PR goes beyond what the issue is about. In general I'd to avoid somewhat mixed updates to avoid even slightest side-effects with the library now used not only in Quarkus but WildFly etc.

Let me offer you a compromise. In this PR lets handle only a case of `EdDSA` conversion, I think a trivial implementation of `fromAlgorithm` can check if it is compares in lower or upper with `EdDSA` and if yes treat it as such, otherwise, just convert to upper case. I'm not too worried at all about cases like `rS256` failing.
And add 2 tests only to address `EdDsa` conversion problem.

Once we merge and tag this fix, lets then do another PR to comprehensively test all the name conversions for all types of algorithms, and there please check with @MikeEdgar if we really want to add this one extra test dependency.

Hope this proposal is not too bad, thanks for understanding.
FYI, the next time I'll be to respond here will be starting from 11th June, talk to you then

Thanks

0rzech:
@sberyozkin The code has already been allowing any-cased letters when signature algorithm is set through properties file. Letters case is already being ignored on that code path and this is already a part of library's public interface.

Only when it is set through header builder method, will it fail for any other letter cases than those matching enum variants, and this is what this PR changes - to match behaviour when reading from file.

The exception is thrown due to how `SignatureAlgorithm` is internally converted to and from String with use of `.getAlgorithm()` method and then `.valueOf(String)`. This is why it can fail even when algorithm is set through builder's `algorithm(SignatureAlgorithm)` method.

So, IMHO, adding a special case just for `EdDSA` will make the code and tests more confusing and the tests won't reflect reality that way.

Nevertheless, I have created #801 with just 3 tests.

implementation/common/pom.xml
0rzech: The whole point of property-based testing is not to rely on one or few specific input examples, but to have tests that cover large space of inputs and help with pinpointing specific errors. `@ParameterizedTest` was not made for this.

IMHO, it's worth investing even more into property-based testing where appropriate, especially in security-related project. For instance, property-based tests could be used to randomize claim headers in JWTs, instead of just making sure that headers work fine for `https://issuer.com` issuer and `customHeader` with `custom-header-value`, because this is what current tests actually prove.

The reason I was hit with the bug was that the tests did not cover the code path my project triggered, that's why I've added tests covering all other signature algorithms. To be fair, though, because all possible algorithms is a relatively small finite input space, `@ParameterizedTest` could do it. It's opposite for invalid inputs in this particular case. 

If you are worried about `jqwik` being some random library, then here's [Oracle's Java Magazine](https://blogs.oracle.com/javamagazine/post/know-for-sure-with-property-based-testing) article about it from almost 5 years ago. Jqwik has also plugins for Spring Boot and Micronaut, and even someone wanted to add jqwik extension to Quarkus, though - unfortunately - to no avail it seems ([here](https://github.com/quarkusio/quarkus/discussions/29834) and [here](https://github.com/quarkusio/quarkus/discussions/29916)), and [here's](https://github.com/apache/kafka/blob/55d38efcc5505a5a1bddb08ba05f4d923f8050f9/raft/src/test/java/org/apache/kafka/raft/internals/RecordsIteratorTest.java) Apache Kafka using `@ParameterizedTest`, `@Property` and normal `@Test` in a single test class no problem.

Having said all that, I fully respect it's your project, not mine. I just hope that at least most arguments are taken into consideration, that's it.

May I ask to help me with understanding why it is such an issue with adding this library, though? I genuinely don't get it, given it's not code that will run in production anyway and the jqwik project seems to have been started over 8 years ago. I may lack some context, though. It's been years since I was programming Java for a living and I'm completely new to Quarkus.

MikeEdgar: I think for me it's less about a new test dependency and more about having a small subset of the tests using a different paradigm. Based on what I've seen/read, I'm not really opposed to something like property-based testing per se, but rather it might be confusing to do things multiple ways within a single project.

Another possible "gotcha" is IDE support for jqwik. It might be fine, but just guessing it might trip up Eclipse's JUnit runner (for example).

0rzech: Thanks for your reply!

You can't use property-based testing only. Not everything makes sense that way, because not every test input is large.

The same way one wouldn't want to do everything with builder pattern, because most constructors have short list of parameters. And the opposite - one wouldn't want to resign from builder pattern completely just to not confuse developers with doing things multiple ways in a single project. It's a matter of familiarity.

I don't know how Eclipse will react, but jqwik is an alternative engine for Junit 5 platform, so there shouldn't be problems with running tests and reporting. Perhaps one would have to make Eclipse aware of jqwik's test annotations, so that the code won't be marked as unused. I don't know, I tend not to write code to make IDE's happy, but to adjust IDEs to the code. ;) Nevertheless, IntelliJ was able to run jqwik tests pretty fine.

sberyozkin:
Closing in favor of #801

PR #801, opened 2024-05-31
Fix IllegalArgumentException when using EdDSA signature algorithm

0rzech:
`JwtClaimsBuilderImpl`.

to `SignatureAlgorithm.valueOf(String)`.

### DIFF
--- a/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
+++ b/implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
@@ -1,5 +1,7 @@
 package io.smallrye.jwt.algorithm;
 
+import java.util.StringJoiner;
+
 /**
  * JWT JSON Web Signature Algorithms.
  *
@@ -20,9 +22,9 @@ public enum SignatureAlgorithm {
     PS384("PS384"),
     PS512("PS512");
 
-    private String algorithmName;
+    private final String algorithmName;
 
-    private SignatureAlgorithm(String algorithmName) {
+    SignatureAlgorithm(String algorithmName) {
         this.algorithmName = algorithmName;
     }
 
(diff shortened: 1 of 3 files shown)


implementation/common/src/main/java/io/smallrye/jwt/algorithm/SignatureAlgorithm.java
```diff
@@ -31,6 +33,19 @@ public String getAlgorithm() {
     }
 
     public static SignatureAlgorithm fromAlgorithm(String algorithmName) {
-        return SignatureAlgorithm.valueOf(algorithmName);
+        try {
+            return SignatureAlgorithm.valueOf(algorithmName.toUpperCase());
```
0rzech, line 37: This gives more opportunities for library users to write whatever case they prefer. OTOH, doing
```java
if (EDDSA.name().equalsIgnoreCase(algorithmName)) {
    return EDDSA;
}
return SignatureAlgorithm.valueOf(algorithmName);
```
instead would introduce compatibility-breaking change.

sberyozkin: @0rzech Thanks for another PR, can you clarify please what a compatibility-breaking change you have in mind ?

I was thinking about a tiny optimization instead:
```
Sttring algorithmNameUpperCase = algorithmName.toUpperCase();
if (EDDSA.name().equals(algorithmNameUpperCase)) {
    return EDDSA;
}
return SignatureAlgorithm.valueOf(algorithmName);
```

but I'm not sure where is the breaking change ?
 

As I said before, I don't think we need to support cases with random low and upper case characters in the algorithm names, `rS256` or `Rs256`, and another set of options of `EdDSA`.

sberyozkin: Never mind, your code looks good,  back from pto, so a bit jet lagged

0rzech: Yeah, I ignored the NPE, because the catch clause will catch it anyway. All in all, `null` algorithm name is just another variant of invalid name. ;)

sberyozkin (dismissed):
@0rzech Thanks for your help and patience with fixing this issue.

Before I merge, can you give me a favor and squash commits, and have `IllegalStateException` thrown indirectly by adding it to https://github.com/smallrye/smallrye-jwt/blob/main/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/impl/ImplMessages.java instead

Thanks

sberyozkin:
@0rzech You may still have to make these two latest tests conditional as Java 11 is still required, let me run the build

sberyozkin:
Looks like so

sberyozkin (approved):

PR #815, opened 2024-08-15
Add smallrye.jwt.verify.secretkey property for inlining secret keys

sberyozkin:
It will make it possible to use env vars to pass the signing secret key information, as asked for in https://github.com/quarkusio/quarkus/discussions/42550.

In fact, passing the secret key content using the public key property also works but it is not great asking users to use that property

### DIFF
--- a/doc/modules/ROOT/pages/configuration.adoc
+++ b/doc/modules/ROOT/pages/configuration.adoc
@@ -39,6 +39,7 @@ SmallRye JWT supports many properties which can be used to customize the token p
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
+|smallrye.jwt.verify.secretkey|none|Secret key supplied as a string.
 |smallrye.jwt.verify.key.location|NONE|Location of the verification key which can point to both public and secret keys. Secret keys can only be in the JWK format. Note that 'mp.jwt.verify.publickey.location' will be ignored if this property is set.
 |smallrye.jwt.verify.algorithm|`RS256`|Signature algorithm. Set it to `ES256` to support the Elliptic Curve signature algorithm. This property is deprecated, use `mp.jwt.verify.publickey.algorithm`.
 |smallrye.jwt.verify.key-format|`ANY`|Set this property to a specific key format such as `PEM_KEY`, `PEM_CERTIFICATE`, `JWK` or `JWK_BASE64URL` to optimize the way the verification key is loaded.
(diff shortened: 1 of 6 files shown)


implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -217,6 +218,13 @@ private static JWTAuthContextInfoProvider create(String publicKey,
     @Inject
     @ConfigProperty(name = "mp.jwt.verify.publickey", defaultValue = NONE)
     private String mpJwtPublicKey;
+
+    /**
+     * @since 1.1
+     */
+    @Inject
+    @ConfigProperty(name = "smallrye.jwt.verify.secretkey", defaultValue = NONE)
+    private String mpJwtSecretKey;
```
MikeEdgar, line 227: Suggested adjustment to JavaDoc and field name (non-MP property)
```suggestion
    /**
     * @since 4.5.4
     */
    @Inject
    @ConfigProperty(name = "smallrye.jwt.verify.secretkey", defaultValue = NONE)
    private String jwtSecretKey;
```

doc/modules/ROOT/pages/configuration.adoc
```diff
@@ -39,6 +39,7 @@ SmallRye JWT supports many properties which can be used to customize the token p
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
+|smallrye.jwt.verify.secretkey|none|Secret key supplied as a string.
```
MikeEdgar, line 42: Will any other properties be ignored if this is set?

MikeEdgar (dismissed):
LGTM, just a question and a minor suggestion.

doc/modules/ROOT/pages/configuration.adoc
sberyozkin: @MikeEdgar Hi Mike, right now the order is as follows, if the MP JWT public key is set, it is picked up, otherwise this one is checked and finally, the keys from locations are loaded. At the moment there is no code there, for example, failing the build if someone sets both public and secret keys, in this case the secret key will be ignored.
Let me add some warnings if several properties pointing to the keys are set

MikeEdgar: Thanks, it just occurred to me because `smallrye.jwt.verify.key.location` below mentions that `mp.jwt.verify.publickey.location` is ignored if both are set.

sberyozkin: Hi @MikeEdgar, I've added some logging, keeping the ordering in mind. The secret key will be checked only if the public key is not set,  and in both cases, if other key properties are set, there will be a warning.

That code requires some good cleanup though in general :-).

If you'd like something else to be updated right now, then just let me know, thanks

MikeEdgar: Latest commit looks good to me, thanks @sberyozkin . Nothing more from me :smile:

MikeEdgar (approved):

PR #816, opened 2024-08-16
Remove the @since tag for smallrye.jwt.verify.secretkey

sberyozkin:
Hi Mike, I thought it could cause some confusion since all other `@since` tags are attached to standard properties and refer to the MP JWT spec versions


MikeEdgar (approved):
:+1:

PR #820, opened 2024-08-20
Allow to build JWT from JSON string

### LINKED ISSUE #813: Add JWT build `claimsJson` method which accepts String
Sometimes it can be handy to do Jwt.claims("{somejson}") when the JSON string is available in the tests, etc. But right now only JsonObject, Map, and the pointer to the JSON file are accepted

sberyozkin:
Fixes #813.

This PR adds another method to the JWT builder API to simplify signing JSON strings.
For example, I've that often enough, in a test I have a JSON string ready (or it may be coming from DB) but we only have `Jwt.claims(Map)`, `Jwt.claims(JsonObject)` and `Jwt.claims(String jsonFileLocation)` start options - so I have to convert JSON String to JsonObject myself, which is not great.

### DIFF
--- a/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/Jwt.java
+++ b/implementation/jwt-build/src/main/java/io/smallrye/jwt/build/Jwt.java
@@ -91,6 +91,16 @@ public final class Jwt {
     }
 
     /**
+     * Creates a new instance of {@link JwtClaimsBuilder} from a JSON string.
+     *
+     * @param json JSON string
+     * @return {@link JwtClaimsBuilder}
+     */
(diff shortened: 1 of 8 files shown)


MikeEdgar (approved):
:+1:

sberyozkin:
Thanks @MikeEdgar, I'll do a minor follow up PR to fix those doc typos in all other similar comments :-)

MikeEdgar (approved):

MikeEdgar:
> I'll do a minor follow up PR to fix those doc typos in all other similar comments :-)

Thanks @sberyozkin, no big deal.

PR #822, opened 2024-08-21
Support for multiple signature algorithms

### LINKED ISSUE #754: allow multiple signature algorithms
https://github.com/smallrye/smallrye-jwt/blob/main/implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java#L224

```java
    @Inject
    @ConfigProperty(name = "mp.jwt.verify.publickey.algorithm")
    private Optional<SignatureAlgorithm> mpJwtPublicKeyAlgorithm;
```

This should be `Optional<Set<SignatureAlgorithm>> mpJwtPublicKeyAlgorithms`,  so that JWT tokens can smoothly switch to another algorithm.

sberyozkin:
Fixes #754


sberyozkin:
Had to replace EdDSA test with the EC test as the former does not work on Java 11.
Mike, @MikeEdgar, Roberto, @radcortez, have a look please when you get a chance.

It has been planned for a while, and right now, it is technically a smallrye jwt specific feature where more than one signature algorithm is supported. But it will have to be done at the MP JWT level as well, where a signature algorithm is a single value while the decryption one is a list, which makes it tricky to map it cleanly to the MP JWT Bridge.

implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/DefaultJWTParser.java
```diff
@@ -179,9 +179,9 @@ private JWTAuthContextInfo copyAuthContextInfo() {
 
     private void setSignatureAlgorithmIfNeeded(JWTAuthContextInfo newAuthContextInfo, String algoStart,
             SignatureAlgorithm newAlgo) {
-        SignatureAlgorithm algo = newAuthContextInfo.getSignatureAlgorithm();
-        if (algo == null || !algo.getAlgorithm().startsWith(algoStart)) {
-            newAuthContextInfo.setSignatureAlgorithm(newAlgo);
+        Set<SignatureAlgorithm> algo = newAuthContextInfo.getSignatureAlgorithm();
+        if (algo == null || !algo.stream().anyMatch(s -> s.getAlgorithm().startsWith(algoStart))) {
```
MikeEdgar, line 183: Can `algo` ever be null? It would at least be `Set.of(SignatureAlgorithm.RS256)` if not set, correct?

MikeEdgar (approved):
Looks good, just a couple comments.

sberyozkin:
Thanks @MikeEdgar, will look at your comments soon

sberyozkin:
Hey @MikeEdgar Resolved your suggestions and also tweaked the code along the way which I copied and pasted for this PR :-)

MikeEdgar (approved):

Issue #826, opened 2024-08-28
Use cache when JWKS provider is down


zilinjak:
Hey, 
We are building an app that validated requests, we are using JWKS over HTTP. The problem here is:
1. App starts
2. Download keys from JWKS over HTTP
3. The cache of JWKS is out of date
4. Tries to get JWKS over HTTP
5. The JWKS endpoint is down
6. This means that the smallrye-jwt should try to use old keys from cache, but it ends with an error instead.

Can we make sure that the old keys will be used in case that the keys from OIDC endpoint can't be loaded ?

sberyozkin:
Hi @zilinjak 

So, we need to have a new property to do:

https://bitbucket.org/b_c/jose4j/src/756257eb92352600d5dde08c2b8ed25db13a8952/src/main/java/org/jose4j/jwk/HttpsJwks.java#lines-93

Would be interested to open a PR ?

sberyozkin:
We have `jwksRefreshInterval` [here](https://github.com/smallrye/smallrye-jwt/blob/main/implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java#L456). So another property like `jwksCacheRetainOnErrorDuration`, default 0, I guess also in mins, should be added, and then propagated via `JWTAuthContextInfo` to be set [here](https://github.com/smallrye/smallrye-jwt/blob/main/implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/AbstractKeyLocationResolver.java#L121)

sberyozkin:
Or even lets call it `jwksRetainOnErrorDuration` since `jwksRefreshInterval` does not have `Cache` in its name, it is implied

The-Funk:
@sberyozkin I'm just now seeing this. I use smallrye-jwt from Quarkus and have noticed this runtime issue when configuring the extension according to the Quarkus docs. I have a separate issue opened a few down from this one if you want to close one of them out as duplicate.

I'd be interested to help with this but don't know if I'd have the time to learn the codebase.

jindrichpilar-kosik:
@The-Funk Are you referring to https://github.com/smallrye/smallrye-jwt/issues/779?  It seems like that one is about first request and retry.  That does not seem like a duplicate of this one.

This issue is about long running app and reloading cache content. When this fails we suggest a grace period and use the expired cache. Of course retry would be great too, but is not a replacement.

sberyozkin:
[...]

PR #840, opened 2024-11-15
Add jwksRetainOnErrorDuration

luneo7:
Adds jwksRetainOnErrorDuration so we can also support Jose4J `setRetainCacheOnErrorDuration` and keep the current for a time period, before trying to fetch the cache again, if some error happens during cache refresh.

Handles https://github.com/smallrye/smallrye-jwt/issues/826


sberyozkin:
Hi @luneo7, thanks, believe it or not, I started typing something as well, trying to fix it, but I'm glad you've done it faster as I was just thinking, how exactly am I going to test it...
I'll have a few minor suggestions specific to the issue shortly, but what I'd appreciate if you could add all the other optimizations in another PR, I'm finding it easier to manage

Thanks

implementation/jwt-auth/src/main/java/io/smallrye/jwt/config/JWTAuthContextInfoProvider.java
```diff
@@ -836,10 +846,11 @@ Optional<JWTAuthContextInfo> getOptionalContextInfo() {
         contextInfo.setTokenAge(mpJwtVerifyTokenAge.orElse(null));
         contextInfo.setJwksRefreshInterval(jwksRefreshInterval);
         contextInfo.setForcedJwksRefreshInterval(forcedJwksRefreshInterval);
+        contextInfo.setJwksRetainOnErrorDuration(jwksRetainOnErrorDuration);
         Set<SignatureAlgorithm> resolvedAlgorithm = mpJwtPublicKeyAlgorithm;
         if (signatureAlgorithm.isPresent()) {
             if (signatureAlgorithm.get().getAlgorithm().startsWith("HS")) {
-                if (!NONE.equals(resolvedVerifyKeyLocation) && resolvedVerifyKeyLocation == mpJwtLocation) {
+                if (!NONE.equals(resolvedVerifyKeyLocation) && resolvedVerifyKeyLocation.equals(mpJwtLocation)) {
```
sberyozkin, line 853: I know this looks suspicious and as far as I recall it was already discussed, but it is actually correct. In any case, let's deal with unrelated improvements in another PR

implementation/jwt-auth/src/main/java/io/smallrye/jwt/auth/principal/AbstractKeyLocationResolver.java
```diff
@@ -115,6 +115,7 @@ protected HttpsJwks initializeHttpsJwks(String location)
                     new InetSocketAddress(authContextInfo.getHttpProxyHost(), authContextInfo.getHttpProxyPort())));
         }
         theHttpsJwks.setSimpleHttpGet(httpGet);
+        theHttpsJwks.setRetainCacheOnErrorDuration(authContextInfo.getJwksRetainOnErrorDuration());
```
sberyozkin, line 118:
```suggestion
        theHttpsJwks.setRetainCacheOnErrorDuration(authContextInfo.getJwksRetainOnErrorDuration() * 60);
```

It depends on updating the property description that it is a duration in mins (I don't recall now why it is not Duration - probably because it is too flexible - one can set a year by default by mistake etc)

sberyozkin:
@luneo7 LGTM overall, thanks very much, minor suggestions are proposed, and please follow up later with a standalone dedicated PR, enjoy the weekend

sberyozkin:
Forgot one thing - please add a new property description in the config docs - probably next to the JWK refresh interval property

luneo7:
@sberyozkin done the requested changes =)

sberyozkin (approved):
Thanks for your help, @luneo7

PR #841, opened 2024-11-15
JWT Auth micro optimization

luneo7:
Micro optimizations, pre-sizing collection creation, removing some auto boxing, and also bypassing set -> set -> array to set -> array conversion


sberyozkin (commented):
It looks neat, but I would prefer to delay it until 4.6.2, thanks @luneo7

sberyozkin (approved):

PR #861, opened 2025-05-25
Make it possible to disable adding default claims

### LINKED ISSUE #858: Issue with automatically added claims
When using smallrye-jwt to create JWT token, some claims like 'exp' are added automatically. This behavior causes issues for me and is breaking RFC 7519 specification. As per section 4.1.4 of this specification claim 'exp' is optional therefore there should be a possibility to not add this claim when generating a token. While for most cases this is not an issue, server that I'm trying to send tokens to, does not accept any additional fields other that specified in server documentation. Because of that I cannot use smallrye-jwt to build my JWT tokens.

sberyozkin:
Fixes #858

The issued at (`iat`), expiration time (`exp`) and `jti` claims are currently set by default if they have not already been set, but some token consumers are too sensitive to the presence of extra claims they don't process (even though, once the signature has been verified, the content is secure).

So I've added a property to make it possible to disable the addition of the default properties

### DIFF
--- a/doc/modules/ROOT/pages/generate-jwt.adoc
+++ b/doc/modules/ROOT/pages/generate-jwt.adoc
@@ -154,6 +154,7 @@ Smallrye JWT supports the following properties which can be used to customize th
 |smallrye.jwt.new-token.issuer|none|Token issuer which can be used to set an `iss` (issuer) claim value if this claim has not already been set.
 |smallrye.jwt.new-token.audience|none|Token audience which can be used to set an `aud` (audience) claim value if this claim has not already been set.
 |smallrye.jwt.new-token.override-matching-claims|false|Override the existing `iss` or `aud` claim values if `smallrye.jwt.new-token.issuer` or `smallrye.jwt.new-token.audience` properties are set.
+|smallrye.jwt.new-token.add-default-claims|true|Disable an automatic addition of the `iat` (issued at), `exp` (expiration time) and `jti` (token identifier) claims when such claims have not already been set.
 |smallrye.jwt.keystore.type|`JKS`|This property can be used to customize a keystore type if either `smallrye.jwt.sign.key.location` or `smallrye.jwt.encrypt.key.location` or both of these properties point to a `KeyStore` file. If it is not set then the file name will be checked to determine the keystore type before defaulting to `JKS`.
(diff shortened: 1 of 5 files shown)


implementation/jwt-build/src/test/java/io/smallrye/jwt/build/JwtSignTest.java
```diff
@@ -386,6 +386,31 @@ void signWithShortRSAKeyAndRelaxedValidation() throws Exception {
...
+        JwtBuildConfigSource configSource = getConfigSource();
+        configSource.setAddDefaultClaims(false);
+        try {
+            String jwt = Jwt.claims(Collections.singletonMap("customClaim", "custom-value"))
+                    .sign(keyPair.getPrivate());
+
+            JsonWebSignature jws = getVerifiedJws(jwt, keyPair.getPublic(), true);
+            JwtClaims claims = JwtClaims.parse(jws.getPayload());
+
+            assertEquals(1, claims.getClaimsMap().size());
```
sberyozkin, line 402: In all other tests where these 3 default claims are added, the expected number of claims is set to 4

doc/modules/ROOT/pages/generate-jwt.adoc
```diff
@@ -154,6 +154,7 @@ Smallrye JWT supports the following properties which can be used to customize th
 |smallrye.jwt.new-token.issuer|none|Token issuer which can be used to set an `iss` (issuer) claim value if this claim has not already been set.
 |smallrye.jwt.new-token.audience|none|Token audience which can be used to set an `aud` (audience) claim value if this claim has not already been set.
 |smallrye.jwt.new-token.override-matching-claims|false|Override the existing `iss` or `aud` claim values if `smallrye.jwt.new-token.issuer` or `smallrye.jwt.new-token.audience` properties are set.
+|smallrye.jwt.new-token.add-default-claims|true|Disable an automatic addition of the issued at ("iat"), expiration time ("exp") and token identifier ("jti") claims when such claims have not already been set.
```
MikeEdgar, line 157: Minor - claim names elsewhere use back-ticks rather than quotes.

MikeEdgar (approved):

MikeEdgar (approved):

PR #914, opened 2026-07-06
Add missing @Inject annotations to JWTAuthContextInfoProvider fields

lrasku:
Absent these, these fields will not get populated from the configuration in a CDI environment.


sberyozkin:
@lrasku That is usually managed by the integrating container like Quarkus or Wildfly

lrasku:
I'm using smallrye-jwt in a Jetty+Weld configuration. But is there something special about these fields that they should *not* be populated via standard CDI injection? This class has many other `@ConfigProperty(name = "smallrye.jwt.XXXX")` fields, but only the ones in this pull request lacked the `@Inject` annotation.

Ladicek:
An injection point without `@Inject` but with a qualifier is automatically recognized by ArC (see https://quarkus.io/guides/cdi-reference#qualified-injected-fields). It is very much non-standard behavior and I'd recommend to _not_ rely on this outside Quarkus.

